PR #432, opened 2015-01-15
Fixed autocomplete if translated, get type from angular

ulai:
The api could not handle search type ('article', 'page') if the site is translated i.e. ('beitrag','seite') in german. So I took the type from angular scope, where it's not translated.


brianhyder:
Great catch and thank you for submitting the patch.

Issue #41, opened 2014-07-06
[PluginFramework] Plugins do not specify what versions of PB they are compatible with


brianhyder:

brianhyder:
Plugins should be able to specify what versions of PB they are compatible with by specifying a property in details.json.  

If a property is not specified it should be assumed to work with any version.  
When a property is specified it should be validated as a valid semantic version expression by the validation code in the plugin validator.  
A check should exist at startup to ensure that when a plugin is being initialized at startup it is still compatible with the current version of PB.

brianhyder:
The UI should not allow for a plugin to be installed if the version is not compatible.  
The install plugin API call should send a 400 status when an attempt to install a plugin is made that is not compatible with the current system core.

JahlomP:
can it be closed?

PR #449, opened 2015-01-25
enhancement for #41

JahlomP:
First version of the solution.
It checks if pbVersion field has been provided by the plugin. If not provided it warns you but assumes the plugin is compatible with any pb version. If provided it validates the semantics of the version provided.


include/service/entities/plugin_service.js
```diff
@@ -1575,6 +1586,13 @@ PluginService.validateDetails = function(details, pluginDirName, cb) {
 	if (!v.validateVersionNum(details.version, true)) {
 		errors.push("An invalid version number ["+details.version+"] was provided.  Must match the form: xx.xx.xx");
 	}
+    
+    //validate pbVersion
+    if(details.pbVersion){        
+        if (!v.validateVersionNum(details.pbVersion, true)) {
+            errors.push("An invalid version number ["+details.pbVersion+"] was provided.  Must match the form: xx.xx.xx");
+        }
+    }
```
brianhyder: you can simplify the expression a bit because the version number is not required.
```
if (!v.validateVersionNum(details.pbVersion, false)) {
  errors.push("An invalid version expression ["+details.pbVersion+"] was provided.");
}
```
You do not need the outer ```if (details.pbVersion)``` because we specify in the validation call that it is not required.

brianhyder:
This is a good start.  There are a few things we should introduce to make sure we validate the semantic versions correctly.  If you aren't familiar with the concept [this](http://semver.org/) is a good read.  We'll want to support the ability for plugins to provide an expression for which versions of PB they are compatible with.  We don't want to write this code ourselves so we'll use a third party library to do the heavy lifting.  So here is what I think we need to do to wrap this up.

1) Introduce ```semver``` (4.2.0).  It should be added into the package.json file for PB.
2) Add a function ```isVersionExpression``` to the ```ValidationService```:
```
//dependencies
var semver = require('semver');

ValidationService.isVersionExpression = function(expression, required) {
  if (!expression && !required) {
    return true;
  }
  return semver.valid(expression) !== null;
};
```
3) In the task you created for the ```pluginInit``` function, when the ```pbVersion``` is provided a call to verify that the current PB version matches the plugin's expression:
```
//should return true
//see https://www.npmjs.com/package/semver for more info on version expressions
semver.satisfies(pb.config.version, details.pbVersion);
```
4) Update the plugin details file's validation task you created to call the new validation function you created.
5) Update the UI to ensure that a plugin with an incompatible version expression cannot be installed.

JahlomP:
thanks

JahlomP:
implemented 1-4. Please verify

brianhyder:
Looks good so far!

JahlomP:
implemented task 5

brianhyder:
Awesome.  I didn't get a chance to look over it today.  I should have some time tomorrow.

JahlomP:
ok

JahlomP:
After your review, I will change all `pbVersion` references to `pb_version`

include/service/entities/plugin_service.js
```diff
@@ -1575,6 +1590,19 @@ PluginService.validateDetails = function(details, pluginDirName, cb) {
...
+    if (!v.isVersionExpression(details.pbVersion, false)) {
+      errors.push("An invalid version expression ["+details.pbVersion+"] was provided.");
+    }
+
+    //validate pbVersion in config against pb version
+    if(details.pbVersion){
+    	if(!semver.satisfies(details.pbVersion, pb.config.version)) {
+      		errors.push("The version of ["+details.name+"] ("+ details.pbVersion +") does not match the pencilblue version (" + pb.config.version + ") you are using.");
+    	}
+    }
```
brianhyder: This block should be updated to: 
```
//validate pbVersion
    if (pb.utils.isString(details.pbVersion)) {
        
        if (!v.isVersionExpression(details.pbVersion, true)) {
            errors.push('An invalid version expression "'+details.pbVersion+'" was provided.');
        }

        //validate pbVersion in config against pb version
        else if(!semver.satisfies(pb.config.version, details.pbVersion)) {
            errors.push("Version "+details.version+" is incompatible with PencilBlue version "+pb.config.version);
        }
    }
```

These changes are necessary because we want to verify that a value exists before do any other tests.  Next, the second if statement becomes an "else if" because if the first error's we know the second will fail.  The next change we make is to the call to satisfies.  The two arguments are in reverse order.

include/validation/validation_service.js
```diff
@@ -151,6 +154,20 @@ ValidationService.validateVersionNum = function(value, required) {
...
+ *
+ * @method isVersionExpression
+ * @param {String} expression
+ * @param {Boolean} required
+ */
+ValidationService.isVersionExpression = function(expression, required) {
+  if (!expression && !required) {
+    return true;
+  }
+  return semver.valid(expression) !== null;
```
brianhyder: How did you test this?   This failed right away when I began testing.  My initial understanding was that the "valid" function validated an expression.  That assumption was incorrect.  The "valid" function only validates a version.  In order to validate the expression we need to update the line to: ```return semver.validRange(expression) !== null;```

brianhyder:
@JahlomP I've added a few comments as to what is needed to be updated.

JahlomP:
Review

brianhyder:
Looks good.  Thanks for the contribution.

JahlomP:
learnt a lot.

Issue #301, opened 2014-10-03
Add an update() function to ArticleService


iv597:
Could be implemented in one of two fashions:

Single update:

    var AS = new pb.ArticleService();
    fields = {$inc: { viewCount: 1 }};
    AS.update(article._id, fields, function(...) {});

Alternatively, bulk update:

    var AS = new pb.ArticleService();
    AS.find(..., function(_, articles) {
        fields = { $set: { myCustomColumn: "Some generic value for each article" } };
        AS.update(articles, fields, function(...) {});
    });

Pardon the probably not dead-on AS.find() usage in that example, you get the idea.

PR #454, opened 2015-02-02
enhancement for #301

JahlomP:
enhancement for #301


include/service/entities/article_service.js
```diff
@@ -191,6 +191,28 @@ ArticleService.prototype.find = function(where, options, cb) {
...
 
 /**
+ * Updates articles
+ *
+ * @param {String} articleId	id of article
+ * @param {Object} options	
+ * @param {Object} fields	fields to update	
+ * @param {Function} cb      Callback function
+ */
+ArticleService.prototype.update = function(articleId, options, fields, cb) {
```
brianhyder: switch the order of options and fields.  Options is usually not required so having it right before the callback allows us to shorten the logic we need to verify that it was provided.

brianhyder: You also need to verify that fields is an actual object since it is required.  You can use the ```pb.utils.isObject(fields)``` function to do that check.  If it isn't an object just ```return cb(new Error('The fields parameter is required'));```

include/service/entities/article_service.js
```diff
@@ -191,6 +191,28 @@ ArticleService.prototype.find = function(where, options, cb) {
...
+	var where = {
+		_id: articleId
+	}
```
brianhyder: This where clause is common across the code base.  For convenience, you can shortcut the syntax with: ```pb.DAO.getIdWhere(articleId)```

include/service/entities/article_service.js
JahlomP: k

include/service/entities/article_service.js
```diff
@@ -191,6 +191,28 @@ ArticleService.prototype.find = function(where, options, cb) {
...
+	
+	var dao  = new pb.DAO();
+	dao.updateFields('article', where, fields, options, function(err, result) {
```
brianhyder: The article service has the ability to operate on articles and pages.  The type can be set for each instance of the service.  Instead of hard coding ```'article'```` use ```this.getContentType()```.

include/service/entities/article_service.js
```diff
@@ -191,6 +191,28 @@ ArticleService.prototype.find = function(where, options, cb) {
...
+	   	if (util.isError(err)) {
+			return cb(err, false);
+		}
+	   	return cb(null, result)
```
brianhyder: Since you aren't changing the result it is safe to pass the cb directly to ```dao.updateFields``` such that your code would look like:
```
dao.updateFields(this.getContentType(), where, fields, options, cb);
```

JahlomP:
Review.

include/service/entities/article_service.js
```diff
@@ -191,6 +191,28 @@ ArticleService.prototype.find = function(where, options, cb) {
...
+ * @param {Function} cb      Callback function
+ */
+ArticleService.prototype.update = function(articleId, fields, options, cb) {
+        if(!pb.utils.isObject(fields)){
+                return cb(new Error('The fields parameter is required'));
+        }
+
+	var where = {
+		_id: pb.DAO.getIdWhere(articleId)
+	}
```
brianhyder: This piece should be:
```
var where = pb.DAO.getIdWhere(articleId);
```

Please test the code before asking for review.

JahlomP: Sorry. I will next time.

Issue #475, opened 2015-02-19
Page Url Characters 404


alexcurtis:
When creating a page url with a non-alphanumeric character such as '-' like: about-us.
returns a 404 when navigating to that page.
No error in the console log.

blakecallens:
Confirmed this is only with the default pencilblue theme. Portfolio works fine.

blakecallens:
Looks like the issue is that the custom URL qualifies as a valid ObjectID, so the controller is trying to load it as such and coming up with no results. Figuring out a fix.

alexcurtis:
@blakecallens Thanks for the quick response.

alexcurtis:
Not sure if this is related but in the portfolio theme.
If I create a page called 'our-history' and setup the navigation to point to that page.
The url snaps to the object id rather than the nice url => /page/54e5e7fbf9e59cb243020971

blakecallens:
@alexcurtis page and article navigation objects are set to point to the permalink of the article or page. That way, if you change the custom URL, the navigation won't break. If you want to use the custom URL, create a *link* navigation item.

This error is fixed in branch PB475. I'll be putting a pull request in for @brianhyder to review before it makes its way to master.

PR #476, opened 2015-02-19
#475 fixed custom URL confused as ObjectID

blakecallens:


plugins/pencilblue/controllers/article.js
```diff
@@ -33,11 +33,9 @@ Article.prototype.render = function(cb) {
 	var custUrl = this.pathVars.customUrl;
 
 	//check for object ID as the custom URL
-	var doRedirect = false;
-	var where      = null;
+	var where  = null;
 	try {
```
brianhyder: This is a lot like the portfolio bug we fixed.  We need to do the validation check on the ```custUrl``` variable instead of the try catch block.  That will allow us to determine if it is an article ID or custom URL.

plugins/pencilblue/controllers/article.js
```diff
@@ -33,11 +33,9 @@ Article.prototype.render = function(cb) {
...
-		where      = {_id: pb.DAO.getObjectID(custUrl)};
-		doRedirect = true;
+		where = {_id: pb.DAO.getObjectID(custUrl)};
```
brianhyder: for this expression we need to abstract away the ```_id``` piece.  

Try this:
```
where = pb.DAO.getIDWhere(custUrl);
```

plugins/pencilblue/controllers/article.js
blakecallens: Right, but keep in mind that the bug here is that the custom URL is being validated as an ObjectID. It's a cosmetic change, more than anything.

blakecallens:
Change is up

brianhyder:
Awesome!

PR #477, opened 2015-02-19
Infinite Scroll - Topic Typo + Support For Offset + Limit From Controller

alexcurtis:
* infiniteScrollTopic typo fixed.
* Added infiniteScrollOffset and infiniteScrollLimit so you can set the current offset + limit from the controller from the get go. Useful when pulling down articles that have topic restrictions.
* Tidied up the parameter construction.

### DIFF
--- a/public/js/infinite_article_scroll.js
+++ b/public/js/infinite_article_scroll.js
@@ -36,14 +36,14 @@ $(document).ready(function()
         if(scrollPosition >= documentHeight - 200)
         {
             self.loadingArticles = true;
-            var parameters = [];  
+            var parameters = [];
 
             //Offset
             if(!self.offset && typeof infiniteScrollOffset !== 'undefined'){
                 self.offset = infiniteScrollOffset;
             }
             if(self.offset){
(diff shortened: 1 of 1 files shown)


blakecallens:
Thanks for improving this! I have one suggestion for the parameter construction that I'll add to the inline comments.

public/js/infinite_article_scroll.js
```diff
@@ -36,20 +36,47 @@ $(document).ready(function()
...
+            
+            //Section or Topic
+            if(typeof infiniteScrollSection !== 'undefined'){
+                parameters.push({section: infiniteScrollSection });
+            }
+            else if(typeof infiniteScrollTopic !== 'undefined'){
+                parameters.push({topic: infiniteScrollTopic });
+            }
+
+            var query;
```
blakecallens: Since this runs off of jQuery, I suggest we use the [param() function](http://api.jquery.com/jquery.param/) to build the string.

alexcurtis:
Sounds like a good idea. I will make the modification.

Issue #481, opened 2015-03-01
Test email not sent


blakecallens:
The API call for sending a test email is not properly converted to Angular.

PR #482, opened 2015-03-01
#481 refactored test email API call

blakecallens:


plugins/pencilblue/controllers/api/admin/site_settings/email/send_test.js
```diff
@@ -24,26 +24,34 @@ function SendTestEmail(){}
...
-
-    var options = {
-        to: post.email,
-        subject: 'Test email from PencilBlue',
-        layout: 'This is a successful test email from the PencilBlue system.',
-    };
-    pb.email.sendFromLayout(options, function(err, response) {
-        if(err) {
-            cb({content: pb.BaseController.apiResponse(pb.BaseController.API_FAILURE, JSON.stringify(err))});
+    this.getJSONPostParams(function(err, post) {
```
brianhyder: In routes.js if you change the block that references this controller to:
```
{
            method: 'post',
            path: "/api/admin/site_settings/email/send_test",
            auth_required: true,
            access_level: pb.SecurityService.ACCESS_ADMINISTRATOR,
            controller: path.join(pb.config.docRoot, 'plugins', 'pencilblue', 'controllers', 'api', 'admin', 'site_settings', 'email', 'send_test.js'),
            content_type: 'application/json',
            request_body: ['application/json']
}
```

You won't need to call ```getJSONPostParams```.  Instead, you just need to call ```var post = this.body```.

blakecallens: Is this something that can be added to the getRoutes function in plugins too? I want to keep this controller in line with what plugins need, so users can use our work as a reference.

brianhyder: Yes, the sample plugin is designed this way so it can be the reference
while our controllers are being converted.

PR #487, opened 2015-03-06
Skip date field validation if empty

jclausen:
Empty date fields should not prevent submission (e.g. - when there is a publish up/publish down date setting for a custom object)

### DIFF
--- a/public/js/angular/filters/parsable_date.js
+++ b/public/js/angular/filters/parsable_date.js
@@ -1,6 +1,8 @@
 angular.module('parseDate', [])
 .filter('parsableDate', function() {
     return function(input) {
+        if(typeof(input)==='undefined')
+            return '';
         var csv = input.split('-').join(',')
                        .split(' ').join(',')
                        .split(':').join(',');


blakecallens:
Thanks for pointing this out. I agree with the intent, but the proposed change could result in JS errors.

The parsableDate filter takes the readable date from the text input part of the date selector and turns it into an ISO 8601 datetime string that can be parsed by JS Date in all browsers. This is used by all date inputs in the admin section, including forms where a date is required.

The place to make the change for custom objects would be in the [angular controller for object_form](https://github.com/pencilblue/pencilblue/blob/master/plugins/pencilblue/templates/angular/admin/content/objects/object_form.html#L92).

blakecallens:
Sorry, I totally read the commit backwards and thought you took out protection instead of added it. Merged.

public/js/angular/filters/parsable_date.js
```diff
@@ -1,6 +1,8 @@
 angular.module('parseDate', [])
 .filter('parsableDate', function() {
     return function(input) {
+        if(typeof(input)==='undefined')
```
blakecallens: For future reference, we wrap all conditionals in brackets, I'll make the change in master.

jclausen:
No worries. Thanks for the fast turnaround.  I'll make sure to wrap with brackets for any future pull requests.  -J

PR #491, opened 2015-03-11
0.4.0 blake

brianhyder:


include/access_management.js
```diff
@@ -131,14 +131,14 @@ module.exports = function(pb) {
...
-     * 
+     *
      * @static
      * @method getRoleToDisplayNameMap
      * @param {Localization} ls
      * @return {Object}
      */
     SecurityService.getRoleToDisplayNameMap = function(ls) {
-        if (util.isFunction(ls)) {
+        if (util.isFunction(ls.get)) {
```
brianhyder: Why did this get changed?  There were some instances where the localization service wasn't getting passed.  The if check is there to provide support for backward compatiablity.  This change will cause a null pointer error in those cases.  Please change it back to ```if (util.isFunction(ls)) {```

include/access_management.js
```diff
@@ -147,7 +147,8 @@ module.exports = function(pb) {
                 'ACCESS_ADMINISTRATOR': ls.get('ACCESS_ADMINISTRATOR'),
             };
         }
-        return null;
+        // Return an empty object instead of null, so things won't break after return
+        return {};
```
brianhyder: The null is here so that the code explicitly has to handle cases where it is calling the function in a deprecated manner.  Was there a specific use case where it needed the change?

include/dao/dao.js
```diff
@@ -809,7 +809,7 @@ module.exports = function DAOModule(pb) {
      * @param {String} oid Object Id String
      * @return {Object}    ObjectID object
      */
-    DAO.getObjectID = function(oid) {
+    DAO.getObjectId = function(oid) {
         pb.log.warn('DAO: getObjectID is deprecated. Use getObjectId instead');
         return DAO.getObjectId(oid);
     };
```
brianhyder: This is the deprecated function.  It must remain cased ```DAO.getObjectID```.  There is already a function cased the appropriate manner.

include/service/entities/user_service.js
```diff
@@ -136,31 +136,50 @@ module.exports = function UserServiceModule(pb) {
...
      * @param {String} currId The Id to be excluded from the list.
      * @param {Function} cb A callback that takes two parameters.  The first is an
      * error, if exists, the second is an array of objects that represent the
      * editor select list.
      */
-    UserService.prototype.getEditorSelectList = function(currId, cb) {
+    UserService.prototype.getEditorSelectList(currId, cb) {
+        pb.log.warn('UserService: getEditorSelectList is deprecated. Use getWriterOrEditorSelectList instead');
+        this.getWriterOrEditorSelectList(currId, cb);
+    };
```
brianhyder: I added the function signature back.  I did this because we don't know if someone was already using it.  I'd like to a take a deprecation strategy where possible to reduce churn on our users.

include/service/entities/user_service.js
```diff
@@ -136,31 +136,50 @@ module.exports = function UserServiceModule(pb) {
...
+     * @param {Boolean} [getWriters=false] Whether to retrieve all writers or just editors.
+     * @param {Function} cb A callback that takes two parameters.  The first is an
+     * error, if exists, the second is an array of objects that represent the
+     * editor select list.
+     */
+    UserService.prototype.getWriterOrEditorSelectList = function(currId, getWriters, cb) {
+        if (util.isFunction(getWriters)) {
+            cb = getWriters;
+            getWriters = false;
+        }
```
brianhyder: I also make ```getWriters``` an optional field and updated the documentation for the parameter to reflect the change

include/service/entities/user_service.js
```diff
@@ -136,31 +136,50 @@ module.exports = function UserServiceModule(pb) {
...
+        
         var self = this;
 
         var opts = {
             select: {
-                _id: 1,
```
brianhyder: I removed the reference to ```_id``` for two reasons.  

1) It is implicitly implied that the id field will be returned by the select 
2) I hate the leading underscore

include/validation/validation_service.js
```diff
@@ -111,7 +111,7 @@ module.exports = function ValidationModule(pb) {
             return true;
         }
 
-        var id = pb.DAO.getObjectID(val);
+        var id = pb.DAO.getObjectId(val);
```
brianhyder: +1 for helping to get rid of all of the deprecation warnings.

plugins/pencilblue/controllers/actions/admin/content/articles/edit_article.js
```diff
@@ -55,7 +55,9 @@ module.exports = function(pb) {
                 }
 
                 //TODO should we keep track of contributors (users who edit)?
```
brianhyder: If you wanted to put this in I don't think it would be hard.  Just create a property ```contributors``` and make it a hash where each key is the user's ID.

plugins/pencilblue/controllers/admin/content/articles/article_form.js
```diff
@@ -44,23 +44,43 @@ module.exports = function(pb) {
...
-                self.onTemplateRetrieved('' + data, function(err, data) {
-                    var result = '' + data;
-                    self.checkForFormRefill(result, function(newResult) {
-                        result = newResult;
-                        cb({content: result});
-                    });
-                });
-            });
+            if(!self.article.author) {
+              self.article.author = self.session.authentication.user._id.toString();
```
brianhyder: It is uncomfortably verbose but we need to refer to the ```_id``` paramter via:

```
self.session.authentication.user[pb.DAO.getIdField()].toString()
```

This helps reduce churn to support multiple DBs in the future.

plugins/pencilblue/controllers/admin/content/articles/article_form.js
```diff
@@ -72,7 +92,7 @@ module.exports = function(pb) {
 
             for(i = 0; i < data.article.article_media.length; i++) {
                 for(j = 0; j < data.media.length; j++) {
-                    if(pb.DAO.areIdsEqual(data.media[j][pb.DAO.idField()], data.article.article_media[i])) {
+                    if(pb.DAO.areIdsEqual(data.media[j][pb.DAO.getIdField()], data.article.article_media[i])) {
```
brianhyder: good catch.  I botched that.

plugins/pencilblue/controllers/api/admin/site_settings/email/send_test.js
```diff
@@ -16,37 +16,44 @@
...
-        var options = {
-            to: post.email,
-            subject: 'Test email from PencilBlue',
-            layout: 'This is a successful test email from the PencilBlue system.',
-        };
-        var emailService = new pb.EmailService();
-        emailService.sendFromLayout(options, function(err, response) {
-            if(err) {
-                cb({content: pb.BaseController.apiResponse(pb.BaseController.API_FAILURE, JSON.stringify(err))});
+        this.getJSONPostParams(function(err, post) {
```
brianhyder: Don't forget that you can declare the incoming content type in the route definition and just call ```this.body```.  It will save you a lot of space and reduce complexity in the controllers

public/js/angular/filters/parsable_date.js
```diff
@@ -1,21 +1,25 @@
...
+      var offsetString = (offset < 10 ? '0' + offset : offset.toString()) + ':00';
+      if(offset < 0) {
+          offsetString = '+' + offsetString;
+      }
+      else {
+          offsetString = '-' + offsetString;
+      }
+
+      return dateArray[2] + '-' + dateArray[0] + '-' + dateArray[1] + 'T' + dateArray[3] + ':' + dateArray[4] + offsetString;
     }
```
brianhyder: Is it time to go to ```moment.js```?

include/access_management.js
blakecallens: This was returning false in my tests, so I moved it to the get function and it worked.

include/access_management.js
blakecallens: The code wasn't handling a null return, it was going straight to acting upon an object.

include/dao/dao.js
blakecallens: Will change.

plugins/pencilblue/controllers/actions/admin/content/articles/edit_article.js
blakecallens: This could be part of our versioning plan.

plugins/pencilblue/controllers/api/admin/site_settings/email/send_test.js
blakecallens: We should go through all the action controllers and change this as well, then.

public/js/angular/filters/parsable_date.js
blakecallens: I say no, because there is only one format we're parsing here. I made a fix to master yesterday that I have to merge in here.

include/access_management.js
brianhyder: What code exactly?  That code that calls should be providing the Localization instance.  Returning null was intentional to highlight code that used the call incorrectly.

include/access_management.js
brianhyder: K

plugins/pencilblue/controllers/api/admin/site_settings/email/send_test.js
brianhyder: Agree.  This is something I'd like to queue up for Q2 is to refactor our controllers to better align with how we constructed the sample controllers.

public/js/angular/filters/parsable_date.js
blakecallens: https://github.com/taylorhakes/fecha

PR #511, opened 2015-03-27
Adding MIME types for fonts and fixing public route QSVars

silverelizard:
Adding MIME types for otf and ttf fonts to eliminate browser warnings about incorrect MIME types.  Making sure qsvars are valid for public resource routes, currently qsvars cause public resource routes to error.

### DIFF
--- a/plugins/pencilblue/controllers/public.js
+++ b/plugins/pencilblue/controllers/public.js
@@ -52,7 +52,8 @@ PluginPublicContentController.prototype.render = function(cb) {
 
     //serve up the content
 	var resourcePath = path.join(pluginPublicDir, postPluginPath);
-	this.reqHandler.servePublicContent(resourcePath);
+	//remove qsvars before loading files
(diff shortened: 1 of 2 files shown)


plugins/pencilblue/controllers/public.js
```diff
@@ -52,7 +52,8 @@ PluginPublicContentController.prototype.render = function(cb) {
 
     //serve up the content
 	var resourcePath = path.join(pluginPublicDir, postPluginPath);
-	this.reqHandler.servePublicContent(resourcePath);
+	//remove qsvars before loading files
+	this.reqHandler.servePublicContent(resourcePath.split('?')[0]);
```
brianhyder: +1 for comments!

brianhyder:
@silverelizard There was a merge conflict so I manually applied the patch file and pushed up the fix to the 0.4.0 branch.  Thank you for passing this one along!

PR #526, opened 2015-04-03
Flexible Analytics Manager

btidwell:
Allow analytics provider plugins to specify template tag names to allow
templates to specify where on the DOM individual providers are
rendered.  (for example: google tag manager requires a data layer in
the head however, the core tag manager snippet might be best placed in
the body of the document).

Also I will follow this change with this submission of a Google Tag Manager plugin on the PencilBlue Plugins page.

### DIFF
--- a/controllers/base_controller.js
+++ b/controllers/base_controller.js
@@ -115,8 +115,10 @@ module.exports = function BaseControllerModule(pb) {
         this.templateService.registerLocal('localization_script', function(flag, cb) {
             self.requiresClientLocalizationCallback(flag, cb);
         });
-        this.templateService.registerLocal('analytics', function(flag, cb) {
-            pb.AnalyticsManager.onPageRender(self.req, self.session, self.ls, cb);
+        pb.AnalyticsManager.onPageRender(self.req, self.session, self.ls, function(err, results){
+          results.forEach(function(result){
+            self.templateService.registerLocal(result.tag, new pb.TemplateValue(result.snippet, false));
+          });
         });
         this.templateService.registerLocal('wysiwyg', function(flag, cb) {
             var wysiwygId = util.uniqueId();
--- a/controllers/base_controller.js
+++ b/controllers/base_controller.js
@@ -19,6 +19,7 @@
 var url       = require('url');
 var Sanitizer = require('sanitize-html');
 var util      = require('../include/util.js');
+var async     = require('async');
 
 module.exports = function BaseControllerModule(pb) {
     
@@ -115,11 +116,6 @@ module.exports = function BaseControllerModule(pb) {
         this.templateService.registerLocal('localization_script', function(flag, cb) {
             self.requiresClientLocalizationCallback(flag, cb);
         });
-        pb.AnalyticsManager.onPageRender(self.req, self.session, self.ls, function(err, results){
-          results.forEach(function(result){
-            self.templateService.registerLocal(result.tag, new pb.TemplateValue(result.snippet, false));
-          });
-        });
         this.templateService.registerLocal('wysiwyg', function(flag, cb) {
             var wysiwygId = util.uniqueId();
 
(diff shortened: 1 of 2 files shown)


include/system/analytics_manager.js
```diff
@@ -112,14 +128,16 @@ module.exports = function AnalyticsManagerModule(pb) {
      * instance of Localization.  The last is a callback that should be called with 
      * two parameters.  The first is an error, if occurred and the second is raw 
      * HTML string that represents the snippet to be executed by the analytics 
-     * plugin. 
+     * plugin.
+     * @param {String} optional name of template tag where the template service to register your snippet
```
brianhyder: Would you mind updating the documentation to:

```
* @param {String} name
* @param {Function} onPageRendering 
* @param {String} [templateTag='analytics'] name of template tag where the template service to register your snippet
```

This will make things show up correctly when we generate our documentation using YUIDoc

controllers/base_controller.js
```diff
@@ -115,8 +115,10 @@ module.exports = function BaseControllerModule(pb) {
         this.templateService.registerLocal('localization_script', function(flag, cb) {
             self.requiresClientLocalizationCallback(flag, cb);
         });
-        this.templateService.registerLocal('analytics', function(flag, cb) {
-            pb.AnalyticsManager.onPageRender(self.req, self.session, self.ls, cb);
+        pb.AnalyticsManager.onPageRender(self.req, self.session, self.ls, function(err, results){
+          results.forEach(function(result){
+            self.templateService.registerLocal(result.tag, new pb.TemplateValue(result.snippet, false));
+          });
```
brianhyder: I don't think this is the right approach here for two reasons:

1) There is a performance concern because calling the onPageRender every time a base controller is instantiated is expensive and you don't know if it is a view or API controller so it could potentially be wasted resources.
2) You execute asynchronously  but do not wait so there is no guarantee that the onPageRender will be available by the time the page renders.

Potential solution:
I think you would be better off to provide a function that returns a list of tags and then register each tag.  This way you would only call the individual analytic providers if and only if they are encountered in the template.  You could then pass the specific flag/tag to the ```pb.AnalyticsManager.onPageRender``` function to know which specific provider to render.

brianhyder:
Thank you for the making the changes.  I'll be looking back over this day and hopefully merging it in.

controllers/base_controller.js
```diff
@@ -125,7 +123,15 @@ module.exports = function BaseControllerModule(pb) {
...
             });
-        });
+        });      
+        var analyticsTags = pb.AnalyticsManager.getTemplateTags();
+        if(analyticsTags.length > 0){
+          analyticsTags.forEach(function(result){
+            self.templateService.registerLocal(result.templateTag, function(flag, cb){
+              pb.AnalyticsManager.onPageRender(self.req, self.session, self.ls, result.providerName, cb);
+            });
+          });
```
brianhyder: This does avoid the performance issues I mentioned but brings to light another issue with this implementation.  It does not support the ability for multiple providers to be rendered under the same analytics tag.  This is needed for backward compatibility.

For example, if someone were running both the google analytics plugin and the ChartBeat plugin only one would get rendered when the ```^analytics^``` flag is encountered in the templates.  The reason is because in your loop you register a single flag name with a single provider.  

The more I ponder over this use case the more I wonder if the core functionality should be to provide an ```analytics_header``` and ```analytics_footer``` flag and let the plugins themselves choose where to render with the default going to the header.  Then if there are other specific use cases for inputing in the body or only in certain circumstances the plugins can register global flags and render appropriately for that specific purpose.

btidwell:
Hey Brian,  Sorry for not getting back to this sooner.  I think for our use case we can leave it as is for now.  For Google Tag Manager, we will just leave it up to the theme template to provide the variable dataLayer and then we will let the plugin render the core GTM script via the ^analytics^ flag.  Based on the way GTM works we don't want both it and Google Analytics installed side by side as GTM handles GA tracking.

PR #528, opened 2015-04-03
Iss#525#527

vqoph:


public/js/admin/elements/content_search.js
```diff
@@ -20,7 +20,7 @@ $(document).ready(function() {
     input.autocomplete({
         source: function( request, response ) {
             $.ajax({
-                url: "/api/content/search/",
+                url: "/api/content/search",
```
brianhyder: great catch.  Thank you for the fix.

Issue #527, opened 2015-04-03
Unable to use preview function in pencilblue default theme


vqoph:
solved as changing  the name of the function from **PreviewModule** in **PageModule**
in plugins/pencilblue/controllers/page.js

blakecallens:
Great catch. Would you mind putting a pull request in for that change?

vqoph:
it's done, i think ... it's my first github pull request, so if i've done something wrong please tell me ;)

brianhyder:
I think there is still an issue with preview.  See my comments on #523

blakecallens:
#523 was a popup blocked issue, so I'm closing this one.

PR #531, opened 2015-04-04
Issue#527

vqoph:
Solving in articles, I renamed the feed module too
https://github.com/pencilblue/pencilblue/issues/527


vqoph:
it's correcting the misuse of PreviewModule namespace but not solving the problem about articles

Issue #523, opened 2015-04-02
Error on updating article


JahlomP:
I get a server error on updating articles. v0.4.0. Can someone verify and let me know if i'm the only one getting this error or is a bug.

JahlomP:
![untitled](https://cloud.githubusercontent.com/assets/4746511/6981517/42c2b0d8-d9f2-11e4-9fe1-98ce832f8786.png)

I'm getting existing url error when updating but this is the only article in the system.

JahlomP:
@brianhyder @blakecallens  a little help here.

blakecallens:
Not happening on my end with the latest version of master.

brianhyder:
I was able to reproduce.

1) click new article
2) fill in title: "hello"
3) fill in custom URL: "wo-wo"
4) fill in content: "hello world"
5) click "Save Draft"

The draft appears to save.  However, afterwards if I click "Preview" nothing happens.  

6) Click articles in the nav menu
7) Select the article you just drafted
8) Click "Preview"

vqoph:
i've got this in the devtools console : 

    XHR finished loading: POST "http://localhost:8080/actions/admin/content/articles/551f3f49a22eee585d488f65"
    POST http://localhost:8080/actions/admin/content/articles/551f3f49a22eee585d488f65 400 (Bad Request)

and this in the  logfile  

    1214  {"level":"warn","message":"DAO: getNotIDField is deprecated. Use getNotIdField instead","timestamp":"2015-04-04T01:35:01.433Z"}
    1215: {"level":"debug","message":"Response Time: 5ms URL=[POST]/actions/admin/content/articles/551f3f49a22eee585d488f65 CODE=400","timestamp":"2015-04-04T01:35:01.434Z"}

qbanguy:
I'm seeing the same error when creating a new article.

blakecallens:
@brianhyder the reason *Preview* did nothing for me is because the popup was blocked by the browser. We have to do a ```window.open()``` call after the draft is saved, so you might have to allow popups to see it.

@JahlomP I recreated your screenshot and the article saved fine on both the new and edit screens.

@qbanguy the EXISTING_URL error (we need to add a localization key for that) means that another item in the system is already using that URL. Are you sure there's no other page, article, or section using the same URL key?

qbanguy:
[...]

PR #532, opened 2015-04-04
Issue#523

vqoph:
Error on updating article #523

add type detection on error parameter to prevent undefined value to throw 'EXISTING_URL'


plugins/pencilblue/controllers/actions/admin/content/articles/edit_article.js
```diff
@@ -62,7 +62,9 @@ module.exports = function(pb) {
                 pb.DocumentCreator.update(post, article, ['meta_keywords']);
 
                 pb.RequestHandler.urlExists(article.url, post.id, function(error, exists) {
-                    if(error !== null || exists || article.url.indexOf('/admin') === 0) {
+                    var testError = (error !== null && typeof error !== 'undefined');
```
brianhyder: Would you mind changing this to: ```!util.isError(error)```?

Issue #533, opened 2015-04-04
Error with portfolio theme 


vqoph:
I have this when i try to view a page with the portfolio theme

    Whoops! Something unexpected happened.

    TypeError: undefined is not a function
        at BlogFilter.BlogModule.Blog.getNavigation (...plugins/portfolio/controllers/blog.js:414:17)
        at Object.tasks.nav (...plugins/portfolio/controllers/blog.js:221:22)
        at ...node_modules/async/lib/async.js:583:25
        at ...node_modules/async/lib/async.js:125:13
        at Array.forEach (native)
        at _each (...node_modules/async/lib/async.js:46:24)
        at Object.async.each (...node_modules/async/lib/async.js:124:9)
        at _parallel (...node_modules/async/lib/async.js:582:20)
        at Object.async.parallel (...node_modules/async/lib/async.js:598:9)
        at BlogFilter.BlogModule.Blog.gatherData (...plugins/portfolio/controllers/blog.js:241:15)

It's look like a deep issue, i looked at include/theme/top_menu.js and when i trace navigation param i have :

    [ { _id: 551f34ccbde79c194e9138bb,
        type: 'page',
        name: 'test',
        item: '551f307ebde79c194e9138a7',
        keywords: [],
        parent: null,
        object_type: 'section',
        url: '/page/551f307ebde79c194e9138a7',
        link: null,
        editor: null,
        new_tab: null,
        created: Sat Apr 04 2015 02:48:12 GMT+0200 (CEST),
        last_modified: Sat Apr 04 2015 02:48:12 GMT+0200 (CEST) }
 ]

i think  **url: '/page/551f307ebde79c194e9138a7'** should be **url: '/page/legal'** 
but a don't know how to fix this

the mongo file in page collection : 

    {
      "_id": ObjectId("551f307ebde79c194e9138a7"),
      "author": "551f2de6bde79c194e91388f",
      "publish_date": new Date("2015-04-04T02:28:00+0200"),
      "meta_keywords": [
        ""
      ],
      "page_media": [

      ],
      "page_topics": [

      ],
      "headline": "Legal",
      "subheading": "Legal",
      "url": "legal",
      "draft": 0,
      "page_layout": "test test test",
      "object_type": "page",
      "created": "2015-04-04T00:29:50.816Z",
      "last_modified": new Date(1428149795171),
      "hero_image": null,
      "id": "551f307ebde79c194e9138a7",
      "template": "pencilblue|index"
    }

[...]

PR #534, opened 2015-04-04
Error with portfolio theme issue #533

vqoph:
render of pages and articles in portfolio theme


brianhyder:
Great catch. thank you!

PR #536, opened 2015-04-07
use the correct instance variable

apmonroe6:
Fixes a bug where it would fail to load custom media service providers.


brianhyder:
Great catch.  Thank you for the fix!

PR #539, opened 2015-04-09
.editorconfig added

marcellkiss:
I just added .editorconfig for unified whitespace handling, I think it's useful for everybody.

### DIFF
--- a/plugins/wp_import/controllers/import_action.js
+++ b/plugins/wp_import/controllers/import_action.js
@@ -60,7 +60,7 @@ module.exports = function ImportWPActionControllerModule(pb) {
                 }
 
                 wpXMLParse.parse(data.toString(), self.session.authentication.user_id, function(err, users) {
-                    if(err) {
+                    if(util.isError(err)) {
                         self.session.error = err.stack;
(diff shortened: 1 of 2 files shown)


blakecallens:
Good call. I'm good with adding this.

brianhyder:
I have hesitations about adding this in.  While it is important to enforce standards across the core code base having this file sitting at the root could easily conflict with those that fork the project and prefer different settings.  

My other hesitation is that the settings are not enforced natively by the editors.  There is no data that suggests that developers know what it is and how to use it let alone download a plugin that may or may  not exist for their specific IDE.

blakecallens:
People who fork the project can change the .editorconfig file to their liking. It doesn't force anyone to use it, unless they have it turned on in their editor and I don't think anyone will be confused by it. It helps to keep third party submissions meeting our preferences.

Let's sleep on it, but I don't see it as a negative.

marcellkiss:
What I see, it is becoming more and more accepted:
http://www.google.co.hu/trends/explore#q=editorconfig
Webstorm for example automatically looks for it, and as I started to write my code (with tabs) I felt the need for it. I think it's nice to have.

PR #540, opened 2015-04-09
check for property before execute toLowerCase()

marcellkiss:
I don't know how did I cause it (didn't code anything, just clicked around on the admin), but when I visited: 
http://localhost:8080/admin/content/objects/types
I got the failure: 

Whoops! Something unexpected happened.
TypeError: Cannot call method 'toLowerCase' of undefine

What I did is just a really small bugfix, now it works, but it would be interesting to investigate the core problem.


blakecallens:
This would be indicative of a deeper error. All custom objects require a name to save. We should do some deeper digging.

brianhyder:
Rejecting this pull request.  The actual issue is that the validation for a custom object type name was requiring a value to be provided.  I am pushing up the fix for code review now.

PR #546, opened 2015-04-10
fixes #545 read more break error

### LINKED ISSUE #545: Read more break not saved if there is no media in wysiwyg
Add a read more break, but no media in an article and the placeholder html element for the break will not be converted to the directive on save.

blakecallens:


brianhyder:
Looks good!

PR #567, opened 2015-04-23
Fixing sort order bug.

markbaird:
The DAO was looking for "orderBy" property in the options, while the sort order was being passed in a property named "order". Also, several services were building sort order as an array of arrays, while MongoDB sort order should always be an object or array of objects. As soon as I fixed dao.js all those array of arrays started causing errors.

Apparently the natural sort order was always returning items in the order you would expect, so this wasn't very noticeable. However as soon as I added a query to my template to return articles with a given topic the sort order began to appear random, which is what led me to find this bug.

I also removed a stray comma in section_service.js, and I fixed the logging of sort order, limit and offset values in dao.js

### DIFF
--- a/include/dao/dao.js
+++ b/include/dao/dao.js
@@ -322,8 +322,9 @@ module.exports = function DAOModule(pb) {
                 .skip(offset);
 
             //apply sort order
-            if (options.order) {
-                cursor.sort(options.order);
+            var orderBy = options.order || options.orderBy;
+            if (orderBy) {
+                cursor.sort(orderBy);
             }
 
             //apply maximum number of results to return
@@ -335,9 +336,9 @@ module.exports = function DAOModule(pb) {
             if(pb.config.db.query_logging){
                 var query = "DAO: SELECT %j FROM %s.%s WHERE %j";
                 var args = [select, self.dbName, entityType, where];
-                if (typeof options.order !== 'undefined') {
+                if (typeof orderBy !== 'undefined') {
                     query += " ORDER BY %j";
-                    args.push(options.order);
+                    args.push(orderBy);
                 }
                 if (typeof options.limit !== 'undefined') {
                     query += " LIMIT %d, OFFSET %d";
--- a/include/dao/dao.js
+++ b/include/dao/dao.js
@@ -293,7 +293,8 @@ module.exports = function DAOModule(pb) {
      * @param {String} entityType The collection to query
      * @param {Object} [where={}] The where clause
      * @param {Object} [select={}] The fields to project
-     * @param {Array} [orderBy] The ordering
+     * @param {Array} [order] The ordering
+     * @param {Array} [orderBy] The ordering. Parameter orderBy is deprecated, use order instead.
      * @param {Integer} [limit] The maximum number of results to return
      * @param {Integer} [offset] The number of results to skip before returning results.
      * @return {Cursor} The MongoDB cursor that provides the results of the query
--- a/include/dao/dao.js
+++ b/include/dao/dao.js
@@ -322,8 +322,8 @@ module.exports = function DAOModule(pb) {
                 .skip(offset);
 
             //apply sort order
-            if (options.orderBy) {
-                cursor.sort(options.orderBy);
+            if (options.order) {
+                cursor.sort(options.order);
             }
 
             //apply maximum number of results to return
@@ -335,13 +335,13 @@ module.exports = function DAOModule(pb) {
             if(pb.config.db.query_logging){
                 var query = "DAO: SELECT %j FROM %s.%s WHERE %j";
                 var args = [select, self.dbName, entityType, where];
-                if (typeof orderBy !== 'undefined') {
+                if (typeof options.order !== 'undefined') {
                     query += " ORDER BY %j";
-                    args.push(orderBy);
+                    args.push(options.order);
                 }
-                if (typeof limit !== 'undefined') {
+                if (typeof options.limit !== 'undefined') {
                     query += " LIMIT %d, OFFSET %d";
-                    args.push(limit, offset);
+                    args.push(options.limit, offset);
                 }
                 args.unshift(query);
                 pb.log.info(util.format.apply(util, args));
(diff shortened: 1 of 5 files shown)


markbaird:
I just realized this might fix the "Recently added plugins" list on the pencilblue.org site, that isn't displaying in order of "recently added".

markbaird:
Wait on this, I found another issue with it that needs to be fixed.

markbaird:
When I deployed this fix to my site running on OpenShift, which is stuck on MongoDB version 2.4.9, it still wasn't sorting articles correctly. So I looked at the MongoDB documentation and I can't find any examples of using an array to specify the sort parameters. All the examples here are using an object with multiple properties: http://docs.mongodb.org/manual/reference/method/cursor.sort/

I changed all the services to use that method of specifying the sort parameter and tested that change against MongoDB versions 2.4.9 and 3.0.1 and everything appears to be sorting correctly now.

include/service/entities/article_service.js
```diff
@@ -132,11 +132,11 @@ module.exports = function ArticleServiceModule(pb) {
 
         //build out the ordering
         var order;
-        if (util.isArray(options.order)) {
+        if (util.isObject(options.order)) {
             order = options.order;
         }
         else {
-            order = [['publish_date', pb.DAO.DESC], ['created', pb.DAO.DESC]];
+            order = {'publish_date': pb.DAO.DESC, 'created': pb.DAO.DESC};
```
brianhyder: I haven't gone back and looked at the mongo documentation yet but how can this guarantee order when hashes are are guaranteed to provide the hash keys back out in any order?

It was our experience early on that the mongo driver excepted two forms of order.  The array syntax as shown above and the object syntax which we used when there was only 1 field to sort on.  It was enormously confusing which is why we attempted to start moving everything to the array syntax.

include/dao/dao.js
```diff
@@ -335,13 +335,13 @@ module.exports = function DAOModule(pb) {
...
+                if (typeof options.order !== 'undefined') {
                     query += " ORDER BY %j";
-                    args.push(orderBy);
+                    args.push(options.order);
                 }
-                if (typeof limit !== 'undefined') {
+                if (typeof options.limit !== 'undefined') {
                     query += " LIMIT %d, OFFSET %d";
-                    args.push(limit, offset);
+                    args.push(options.limit, offset);
```
brianhyder: good catch.  Thank you for fixing.

include/dao/dao.js
```diff
@@ -322,8 +322,8 @@ module.exports = function DAOModule(pb) {
                 .skip(offset);
 
             //apply sort order
-            if (options.orderBy) {
-                cursor.sort(options.orderBy);
+            if (options.order) {
+                cursor.sort(options.order);
```
brianhyder: I worry about switching the name here at such a low level.  This causes a break change to our API.  I am fine with moving to ```order``` instead of ```orderBy``` but we need to support both attributes or risk causing issues to existing deployments.  Maybe introduce a local variable to represent order: ```var order = options.order || options.orderBy;```

include/service/entities/article_service.js
brianhyder: Ok, part of the issue is mongo's documentation site.  When you click on the ```Cursor``` prototype and then click on the ```sort``` function in the left hand menu it takes you to the ```AggregationCursor``` page and not the generic cursor page.  Here is the correct link for the function:

http://mongodb.github.io/node-mongodb-native/2.0/api/Cursor.html#sort

markbaird: That's weird that they show the [key, direction] format in the documentation. That has to be a typo. Here's what I'm seeing if I run queries directly against MongoDB, outside of PencilBlue:

MongoDB 3.0.1:
db.getCollection('article').find({}).sort(['publish_date', -1])   - Records returned in natural order, not by publish_date
db.getCollection('article').find({}).sort([['publish_date', -1]]) - Error: BadValue bad order array [2]
db.getCollection('article').find({}).sort({'publish_date': -1}) - Records returned in order of publish_date
db.getCollection('article').find({}).sort([{'publish_date': -1}]) - Records returned in order of publish_date

MongoDB 2.4.9:
db.getCollection('article').find({}).sort(['publish_date', -1])   - Records returned in natural order, not by publish_date
db.getCollection('article').find({}).sort([['publish_date', -1]]) - Records returned in natural order, not by publish_date
db.getCollection('article').find({}).sort({'publish_date': -1}) - Records returned in order of publish_date
db.getCollection('article').find({}).sort([{'publish_date': -1}]) - Records returned in natural order, not by publish_date

I agree with you on the hash key order issue. It looks like arrays work on MongoDB 3, but not MongoDB 2.4. Any suggestions on how to proceed?

include/dao/dao.js
markbaird: The change is in _doQuery(). The underscore indicates that it is intended to be a private function only used inside dao.js, correct? I would think any existing deployments would be calling the q() function in dao.js which was setting the "order" property, not "orderBy".

markbaird:
Please review the code now. Sorting should now work with newer versions of MongoDB, and by removing the isArray() check in article_service you can now pass a single sort key into that from your controllers to get sorting working on older versions of MongoDB.

I also added that bit in dao.js you requested, to check for both "order" and "orderBy" properties. Covervalls is complaining about that additional line of code though.

include/dao/dao.js
```diff
@@ -322,8 +322,9 @@ module.exports = function DAOModule(pb) {
                 .skip(offset);
 
             //apply sort order
-            if (options.orderBy) {
-                cursor.sort(options.orderBy);
+            var orderBy = options.order || options.orderBy;
+            if (orderBy) {
+                cursor.sort(orderBy);
```
brianhyder: Awesome, one small nit-picky change.  Would you mind updating the documentation for the function to include the ```order``` option and mark the ```orderBy``` option as deprecated?

brianhyder:
I wouldn't worry about coveralls.  It is always complaining.  We are still working on getting our code coverage numbers up.

include/dao/dao.js
markbaird: Are you guys using JSDoc to generate your documentation? I don't see a way to mark a single parameter as deprecated.

brianhyder: We use Yuidoc.  I would just add a quick note in the description for the parameter.

PR #569, opened 2015-04-25
fixes #467 - changed article/page in nav item form to select

### LINKED ISSUE #467: Improve the UI/UX of the page/article input on new navigation items
This input is confusing to users. We need a simpler way to choose a page or article. The tooltip needs to be fixed too.

blakecallens:
I changed this to a select element, because it puts it in like with the other 1 to 1 selections in the admin section, and will be far less confusing for users. Loading all pages and articles for this screen is not optimal, but this is a very low traffic page and will function fine for the time being.

When we move the admin interface to a ng-resource model, we can lazy load the articles and pages into the select element.


blakecallens:
Also, I found the old "section" folder was still in the templates directory. Not sure if it never got removed or crept back in, but I pulled it out.

plugins/pencilblue/controllers/admin/content/navigation/nav_item_form.js
```diff
@@ -20,147 +20,159 @@ var async = require('async');
...
+
+      types: function(callback) {
+        callback(null, SectionService.getTypes(self.ls));
+      },
+
+      articles: function(callback) {
+        dao.q('article', {where: {draft: {$ne: 1}}}, function(err, articles) {
+          callback(err, articles);
+        });
+      },
```
brianhyder: I really don't like this.  This is a step in the wrong direction no matter how little traffic the page gets.  It is still a burden on the DB to return data for a page that only shows the data if a specific selection is made.  If we need an API endpoint to retrieve and load all articles/pages we can do it.  That is a much better stop gap solution until a better UX can be provided.

blakecallens:
Let me figure out the best way to handle that. I don't like that it's a single use case, where we have to design a whole new process for a single element in a single page.

brianhyder:
I totally get that.  I don't think it will be a one time thing.  I think it will most likely set the stage for what we move to for all of our multi-select type drag and drops items too.  We need to replace those bulk load components with "on-interaction" or "on page load" components to continue to scale and support the growing number of large clients.

Let me know if you want an API end point for this.  My next task is diving into the base object services so I'll need an end-point or two to sufficiently test it.

blakecallens:
I definitely think we'll need a generic API call on the admin side to call
DAO.q with an offset and limit. I'll get requirements together.

markbaird:
Currently, if you create a custom object with a field of type "Peer Object", and then want to create an instance of this and link it to an article, you get a dropdown control containing all the articles in the system. This is a nightmare when you have hundreds of articles in your system.

I like the search control in the navigation admin much better than the dropdown control in the custom object admin. I was actually going to request the search feature be added to the custom object edit page. So it's not a single use case, there are at least 2 places in the admin that could use this.

blakecallens:
Well, let me rethink how we want to approach this then. I think some of the confusion our users have with the search only approach is they can't remember the name of the item they're looking for, so we need some way of browsing too.

I'm going to explore an interface element that works like our one to many drag and drop, which is searchable, but only allows the selection of one item. I'm going to close this pull request, since it looks like there's nothing I can use from it.

PR #575, opened 2015-04-28
Prevent draft articles and pages from showing up in feed and sitemap.

markbaird:
I just noticed a draft article showing up in my RSS feed. Draft articles and pages weren't being filtered out of the RSS feed or the sitemap. This commit fixes that.


brianhyder:
Good catch and thanks for the fix!

PR #581, opened 2015-05-03
fixes #578 - correct comment allow behavior

### LINKED ISSUE #578: Comments can't be allowed/disallowed per article
The "allow comments" option per article doesn't seem to do anything. Only the global "allow comments" option seems to actually restrict comments.

blakecallens:


plugins/pencilblue/controllers/index.js
```diff
@@ -280,7 +280,7 @@ module.exports = function IndexModule(pb) {
         ats.registerLocal('author_position', content.author_position ? content.author_position : '');
         ats.registerLocal('media_body_style', content.media_body_style ? content.media_body_style : '');
         ats.registerLocal('comments', function(flag, cb) {
-            if (isPage || !contentSettings.allow_comments || !content.allow_comments === false) {
+            if (isPage || !contentSettings.allow_comments || !content.allow_comments) {
```
brianhyder: Does this also need to be changed/updated in Portfolio, Jennifer, and DevBlog themes?

PR #586, opened 2015-05-04
article level disallow comments in portfolio theme

kcdodd:
Added logic to disallow comments on a per-article basis to the portfolio
theme


blakecallens:
Thanks for the fix!

PR #598, opened 2015-05-11
Base Object Service

brianhyder:
Provides a common interface for interacting with system objects.  This is the implementation for the base service (abstracted CRUD ops) as well as a TopicService implementation and samples in the sample code plugin.

Also provides a bug fix and performance updates for article rendering


include/service/entities/article_service.js
```diff
@@ -435,88 +461,150 @@ module.exports = function ArticleServiceModule(pb) {
...
+     * keywords - an array of SEO keywords that describe the content, 
+     * thumbnail - a URI path to the thumbnail image 
      */
-    ArticleService.getMetaInfo = function(article, cb)
-    {
-        if(typeof article === 'undefined')
-        {
-            cb('', '', '');
-            return;
+    ArticleService.prototype.getMetaInfo = function(article, cb) {
```
brianhyder: @markbaird This is where the issue was.  You were correct that things were not playing nicely when called in parallel.  I ended up deprecated the function directly off of the prototype and created a prototype function that in turn tasks itself out and executes in parallel.  

The whole exercise helped me find a few performance issues so a big thanks for the find.

PR #599, opened 2015-05-11
Fixing articleService.getMetaInfo() calls to work with the new version.

markbaird:
Updating the signature of the callback passed to getMetaInfo(), in the themes that are using it, to work with the new version.


brianhyder:
I completely botched that.  Thank you

PR #610, opened 2015-05-15
Fix for timestamp settings property duplication

andparker:
The setting value for two digit date was being incorrectly set to the two digit time setting value on duplication of the property. This was also preventing the two digit time setting from being set correctly.


brianhyder:
Good catch

PR #620, opened 2015-05-20
Fixes #612. Corrects issues where the cache entity service was not pr…

### LINKED ISSUE #612: Compatibility issue with Redis Labs
Given the fact that [Redis Labs only allows one database](https://support.redislabs.com/hc/en-us/articles/202215763-Can-I-have-multiple-databases-on-my-Redis-DB-Getting-ERR-invalid-DB-index-), instead of multiple numbered databases, PencilBlue currently fails to connect successfully to Redis Labs.

I found that `redis_registration_provider.js` specifies a Redis database index of `2`, on line 46. RedisLabs only allows an index of `0`, and I found that if I change that to `0` then I seem to be able to use PencilBlue with a Redis Labs database. However I'm worried about what the implications of changing the registry database index are. Does PencilBlue need more than one Redis database in order to work correctly?

brianhyder:
…operly serializing and deserializing values.

Plugin & theme settings now properly cache too.

### DIFF
--- a/include/service/cache_entity_service.js
+++ b/include/service/cache_entity_service.js
@@ -31,14 +31,17 @@ module.exports = function CacheEntityServiceModule(pb) {
      * @module Services
      * @class CacheEntityService
      * @constructor
-     * @param {String} objType
-     * @param {String} valueField
-     * @param {String} keyField
+     * @param {String} [objType]
+     * @param {String} [valueField]
+     * @param {String} [keyField]
+     * @param {Integer} The number of seconds that a value will remain in cache 
+     * before expiry.
      */
-    function CacheEntityService(objType, valueField, keyField){
+    function CacheEntityService(objType, valueField, keyField, timeout){
         this.type       = 'Cache';
         this.objType    = objType;
         this.keyField   = keyField;
+        this.timeout    = timeout || 0;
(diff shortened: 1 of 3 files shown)


include/service/cache_entity_service.js
```diff
@@ -52,16 +55,14 @@ module.exports = function CacheEntityServiceModule(pb) {
     CacheEntityService.prototype.get = function(key, cb){
 
         var self = this;
-        pb.cache.get(key, function(err, result){
+        pb.cache.get(key, function(err, result){console.log(util.inspect(result));
```
markbaird: Did you intend to commit this console.log()?

include/service/cache_entity_service.js
```diff
@@ -110,10 +122,10 @@ module.exports = function CacheEntityServiceModule(pb) {
                 }
                 rawVal[self.valueField] = value;
                 val                     = JSON.stringify(rawVal);
-            }
+            }console.log('TIMEOUT: '+self.timeout);
```
markbaird: Also this one

include/service/cache_entity_service.js
brianhyder: Fail... No, I thought I got them all. Thank you for the heads up.

PR #621, opened 2015-05-25
A more complete base object service

brianhyder:


eugene-chang-cb:
This PR would help us greatly.  We just have a couple of simple concerns:

* The context is not being passed into the specific services, though they do have constructors set up.
* Can a controller ever have more than one specific service? Should we assume that self.service always refers to the one (and only) service the controller can have?

We were wondering if we could add some sort of "virtual" method at the BaseController level, say something like getControllerExecutingContext(), and the base method would return a clone of the base context (which could just be {}); then in BaseController's init, we could call it by doing something like:
`self.context = self.getControllerExecutingContext()`
and in specific controller's init, we could have
`self.service = new TopicService(util.clone(self.context));`
(maybe with the cloning happening somewhere else)

With this implementation, we still cannot pass any form of context into the BaseObjectService without having to modify every controller affected by the refactor.  Could you please consider this suggestion and give us feedback? Thanks!

brianhyder:
@eugene-chang-cb Thank you for the comments.  I have updated per the request to provide convenience functions for creating contexts.  In addition, I have made the changes to the article and topic api controllers to fit the new pattern.  

To address the two bullet above:
It is possible for a controller to leverage more than once service.  For example, the article controller will eventually need a comment service to handle routes (/articles/{:articleId}/comments/{:id}) that allow for managing comments for an article. In the overriden ```init``` function in the article api controller a second service, most likely ```commentService``` would be created.  The base api controller would not need to know about this particular service but the custom route handlers implemented in the article api controller would need access to it.

The ```service``` property on the article and topic api controllers is leveraged by the underlying ```BaseApiController``` not the ```BaseController```.  The controllers that handle views are not abstracted the same way as the api controller.  However, any services that are needed by the controller should still be instantiated in an overriden ```init``` function for the controller just as they are done in the api controller implementations.

Instantiating the services in the ```init``` function makes it easier to mock out the controllers for testing purposes because it isolates them.  

I don't think you will be able to get away from touching every controller in order to complete the refactor.  However, a consistent pattern moving forward (as I hope we have established) will make it easier along the way.  I think this will be especially true as we add other controllers for specific functionality.  You all will not have to go back and modify those as the context is only constructed in a single prototype (```BaseController```).

Let me know if y'all have any other questions otherwise I will merge tomorrow noonish.

eugene-chang-cb:
@brianhyder Thank you for the changes.  They will definitely help us with merging master into our multitenant branch.  And while, yes, I guess we would have to alter every controller slightly (extend from a different base class, perhaps), it would still prevent us from having to examine every controller.

PR #660, opened 2015-06-21
Wysiwyg3.0

brianhyder:


brianhyder:
@blakecallens I'm going to start going through and reviewing.  I'm going to comment everything then we can figure out how to divi up any changes.

plugins/pencilblue/controllers/admin/elements/wysiwyg.js
```diff
@@ -0,0 +1,42 @@
...
+module.exports = function WysiwygModule(pb) {
+
+  //pb dependencies
+  var util = pb.util;
+
+  /**
+  * Route controller for the wysiwyg AngularJS directive.
+  * @class WysiwygController
+  * @constructor
+  */
```
brianhyder: ```@extends BaseController```

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+  }
+
+  $scope.paginate = function(field, paginationIndex) {
+    field.paginationIndex = paginationIndex;
+
+    paginationService.paginate(field.available_objects, field.paginationIndex, field.paginationLimit, function(availableObjects, paginationPages) {
+      field.available_objects = availableObjects;
+      field.paginationPages = paginationPages;
+    })
+  }
```
brianhyder: missing semi-colons

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+      accept: function (sourceItemHandleScope, destSortableScope) {
+        var itemToTest = sourceItemHandleScope.itemScope.object;
+
+        for(var i = 0; i < active.length; i++) {
+          if(itemToTest._id.toString() === active[i]._id.toString()) {
+            return true;
+          }
+        }
+        for(var i = 0; i < inactive.length; i++) {
+          if(itemToTest._id.toString() === inactive[i]._id.toString()) {
```
brianhyder: why do you need to call ```toString``` on each ```_id``` property?

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+    return field.name.toLowerCase().split(' ').join();
+  }
+
+  $scope.setBoolean = function(fieldName, setting) {
+    $scope.customObject[fieldName] = setting;
+  };
+
+  $scope.setDateToNow = function(fieldName) {
+    $scope.customObject[fieldName] = $filter('date')(new Date(), 'MM-dd-yyyy HH:mm');
+  }
```
brianhyder: missing semi-colons on function declarations

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+  $scope.getFieldUID = function(field) {
+    return field.name.toLowerCase().split(' ').join();
+  }
+
+  $scope.setBoolean = function(fieldName, setting) {
+    $scope.customObject[fieldName] = setting;
+  };
+
+  $scope.setDateToNow = function(fieldName) {
+    $scope.customObject[fieldName] = $filter('date')(new Date(), 'MM-dd-yyyy HH:mm');
```
brianhyder: we need to start localizing the date formats.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
-                $scope.objectType.fields[i].paginationIndex = 0;
-                $scope.objectType.fields[i].paginationLimit = 12;
-                $scope.paginate($scope.objectType.fields[i], 0);
-            }
+        index++;
+        self.loadWysiwyg(index);
+      });
+    };
+
+    this.loadWysiwyg(0);
```
brianhyder: This pattern forces us to load one wysiwyg at a time.  We need to change this to do it in parallel to keep page rendering times down.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+    else if($scope.objectType.fields[i].field_type === 'date') {
+      $scope.customObject[$scope.objectType.fields[i].name] = $filter('date')($scope.customObject[$scope.objectType.fields[i].name], 'MM-dd-yyyy HH:mm');
+    }
+    else if($scope.objectType.fields[i].field_type === 'peer_object') {
+      $scope.customObject[$scope.objectType.fields[i].name] = $scope.customObject[$scope.objectType.fields[i].name] ? $scope.customObject[$scope.objectType.fields[i].name]._id.toString() : '';
+    }
+    else if($scope.objectType.fields[i].field_type === 'child_objects') {
+      $scope.objectType.fields[i].searchText = '';
+      $scope.objectType.fields[i].paginationIndex = 0;
+      $scope.objectType.fields[i].paginationLimit = 12;
```
brianhyder: This magic number should be a constant at the top of the file

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+      }
+
+      $('.date_field').datetimepicker({format: 'm-d-Y H:i'});
+    }
+
+    if(scope.$last){
+      for(var i = 0; i < scope.objectType.fields.length; i++) {
+        if(scope.objectType.fields[i].field_type === 'date') {
+          this.testForDateFields();
         }
```
brianhyder: I'm trying to understand what this piece is doing.  Won't this potentially cause line ~198 to fire multiple times?

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+
+          for(var i = 0; i < scope.availableElements.length; i++) {
+            if(scope.availableElements[i].type === block) {
+              return scope.availableElements[i];
+            }
+          }
+
+          scope.setElement('p');
+          return scope.availableElements[0];
+        }
```
brianhyder: missing semi-colon

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
```diff
@@ -1,251 +1,210 @@
...
+      if(!$('.date_field').length) {
+        setTimeout(self.testForDateFields, 100);
+        return;
+      }
+
+      $('.date_field').datetimepicker({format: 'm-d-Y H:i'});
+    }
+
+    if(scope.$last){
+      for(var i = 0; i < scope.objectType.fields.length; i++) {
```
brianhyder: if you really do need to loop over all elements then simplify it by using ```scope.objectType.fields.forEach(...```

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+        };
+
+        scope.showInsertMediaModal = function() {
+          angular.element(element).find('[contenteditable]').focus();
+          scope.saveSelection();
+          angular.element(element).find('[insert-media-modal]').modal('show');
+        };
+
+        scope.associateMedia = function() {
+          angular.element(element).find('[insert-media-modal]').modal('hide');
```
brianhyder: I would make the hard coded strings constants at the top of the file.

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+
+        scope.getMediaFormat = function() {
+          var mediaFormat = '/position:' + scope.wysiwyg.mediaPosition;
+
+          if(scope.wysiwyg.mediaMaxHeight) {
+            mediaFormat = mediaFormat.concat(',maxheight:' + scope.wysiwyg.mediaMaxHeight + scope.wysiwyg.mediaMaxHeightUnit);
+          }
+
+          return mediaFormat;
+        }
```
brianhyder: semi-colon

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+          }
+        };
+
+        scope.getSelection = function() {
+          if(scope.editableSelection) {
+            return rangy.getSelection().toString();
+          }
+
+          return '';
+        }
```
brianhyder: semi-colon

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+          if(scope.wysiwyg.currentView !== 'editable') {
+            return;
+          }
+
+          var index = scope.wysiwyg.layout.indexOf('^media_display_');
+          if(index === -1) {
+            return;
+          }
+
+          var startIndex = index + 15;
```
brianhyder: make the prefix ```^media_display_``` a constant at the top of the file then reference by ".length" rather than magic number.

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+            if(!result.code) {
+              var mediaPreview = result.data;
+
+              scope.wysiwyg.layout = scope.wysiwyg.layout.split('^' + mediaTag + '^').join(mediaPreview);
+              if(scope.wysiwyg.layout.indexOf('^media_display_') > -1) {
+                scope.loadMediaPreviews();
+              }
+            }
+          });
+        }
```
brianhyder: semi-colon

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+              }
+
+              s++;
+              if(s >= readMoreCount) {
+                scope.layout = tempEditable.html();
+              }
+            });
+          };
+
+          this.convertMedia = function() {
```
brianhyder: separate out ```convertMedia``` and ```covertReadMore``` into functions outside of ```scope.setPublicLayout``` for readability.

public/js/angular/directives/wysiwyg.js
```diff
@@ -0,0 +1,405 @@
...
+              for(var j = 0; j < subTags.length; j++) {
+                mediaTags.push('^' + $(subTags[j]).attr('media-tag') + '^')
+              }
+
+              angular.element(this).replaceWith(mediaTags.concat(''));
+
+              i++;
+              if(i >= mediaCount){
+                self.convertReadMore();
+              }
```
brianhyder: Is the .each function synchronous or asynchronous?  If synchronous (I think it is) then move then move the call to ```convertReadMore``` to below the conclusion of the call to ```.each```

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: Performance wise, what would be faster, an object compare or a no-type-cast string compare after the toString() function?

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: we'll need to bring it into global JS scope then.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: angular.forEach would be more in line, I think

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: This is called once after the ng-repeat is done. It triggers the jquery datetimepicker to be applied to the date fields.

public/js/angular/directives/wysiwyg.js
blakecallens: That's just an attribute selector in jquery and is normal syntax. We don't need to make them constants, IMO.

public/js/angular/directives/wysiwyg.js
blakecallens: Synchronous

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
brianhyder: Good question.  I was thinking that coming from the server the ```_id``` property was already serialized to a string.  Did that change after the mongo driver update?

blakecallens: You're right. I'll remove the toString calls.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
brianhyder: For the angular noob, what kind of effort does that entail?  I was thinking that was just injection of a built in service that just had to be configured to point to the localization files.  I'm probably over simplifying it though.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
brianhyder: ah, got it.  Then do you need to break out of the loop after it makes the call?

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: We could make it it's own directive and bring that in, or we could just
make it a variable outside of angular and bring it into scope. This is so
small that I think the first approach would be overkill.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: No, that's what scope.$last is. It'll only run that code on the last run of
the ng-repeat.

plugins/pencilblue/templates/angular/admin/content/objects/object_form.html
blakecallens: Good news is that this function is totally unnecessary. The angular directive handles all this itself. Removing.

blakecallens:
We should be merging this into a 0.4.1 branch instead of master, IMO.

brianhyder:
Declining and opening another PR to move this to the 0.4.1 branch

PR #667, opened 2015-06-30
fixes #580 - user self management converted to angular model

### LINKED ISSUE #580: Migrate user account management to new AngularJS schema
I cannot upload a "User photo" from the personal account manager, but I *can* do it from the admin panel. When I upload a user photo from admin panel it appears where it should. But, if I do anything with the personal account (ie hit the save button), then the user photo is wiped out. Just to reiterate, as it is a user photo has to be re-uploaded from the admin panel anytime I change anything from the personal account manager.

blakecallens:


blakecallens:
Forgot to put this against 0.4.1.

Issue #509, opened 2015-03-27
Allow admin user to set language preference.


blakecallens:
Users might not want to use the language set in their browser.

blakecallens:
It would be better if we changed this to each user having an individual setting.

PR #674, opened 2015-07-05
Feature/509

brianhyder:
Resolves issue #509. Provides the ability to set language preference at any point during the session and from user preference.


blakecallens:
Everything looks good, except that I had to logout and log back in for my language changes to take effect. Is there a way to instantly do it if you're changing your own settings?

brianhyder:
Should work if you do it through the profile view. I'm guessing forgot to
change it for the admin view

blakecallens:
Yes, worked in profile view. I noticed that reader login is broken. I'm going to fix it.

blakecallens:
Change in plan. It's just broken on your feature branch. Can you look into it?

blakecallens:
Ok, another change in plan. It's broken in both, so I'll fix it in 0.4.1 and you can merge the change.

PR #680, opened 2015-07-07
Multi Tenancy

btidwell:
Host multiple sites under one Pencilblue installation.  Merged with changes from 0.4.1.


.gitignore
```diff
@@ -19,7 +20,9 @@ plugins/*
 !plugins/wp_import/**
 **/node_modules
 **/node_modules/**
-config.js
 config.json
+coverage/
```
brianhyder: Why was coverage added twice?

controllers/admin/base_admin_controller.js
```diff
@@ -0,0 +1,75 @@
...
+  "use strict";
+
+  var util = pb.util;
+  var BaseController = pb.BaseController;
+
+  /**
+   * This class serves as a base class for all the controllers used in the admin control panel
+   * @constructor
+   */
+  function BaseAdminController() {}
```
brianhyder: We need to add the following to ensure that the prototype shows up correctly in the auto generated documentation:
```
@class BaseAdminController
@constructor
@extends BaseController
```

controllers/admin/base_admin_controller.js
```diff
@@ -0,0 +1,75 @@
...
+  util.inherits(BaseAdminController, BaseController);
+
+  /**
+   * Initializes the admin controller with site-related info
+   * @override
+   * @param props
+   * @param cb
+   */
+  BaseAdminController.prototype.init = function (props, cb) {
```
brianhyder: With the ```param``` attribute it needs the type declaration for the auto docs.

```
@override
@method init
@param {Object} props
@param {Function} cb
```

controllers/admin/base_admin_controller.js
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Retrieves a context object that contains the necessary information for
+   * service prototypes
+   * @method getServiceContext
+   * @return {Object}
+   */
+  BaseAdminController.prototype.getServiceContext = function(){
+    return util.merge(BaseAdminController.super_.prototype.getServiceContext.apply(this), { onlyThisSite: true});
+  };
```
brianhyder: The base controller is already creating a new context object each time ```getServiceContext``` is called.  It is not necessary to call merge.
```
var context = BaseAdminController.super_.prototype.getServiceContext.apply(this);
context.onlyThisSite = true;
return context;
```

controllers/base_controller.js
```diff
@@ -67,6 +71,17 @@ module.exports = function BaseControllerModule(pb) {
...
+    /**
+     * The prefix in the content-type header that indicates the charset used in 
+     * the encoding
+     * @static
+     * @private
+     * @readonly
+     * @property CHARSET_HEADER_PREFIX
+     * @type {String}
+     */
+    var CHARSET_HEADER_PREFIX = 'charset=';
```
brianhyder: this variable isn't used anywhere in the file and is private.  it should be removed.

controllers/base_controller.js
```diff
@@ -116,54 +132,76 @@ module.exports = function BaseControllerModule(pb) {
...
-            activeTheme: this.activeTheme
+            site: this.site,
+            activeTheme: this.activeTheme,
+            onlyThisSite: true
         };
 
-        cb();
+        var siteService = new SiteService();
+        siteService.getByUid(this.site, function (err, siteInfo) {
+            if(pb.util.isError(err)) {
```
brianhyder: util is already required.  No need to call ```pb.util```.  Just use ```util```

controllers/base_controller.js
```diff
@@ -116,54 +132,76 @@ module.exports = function BaseControllerModule(pb) {
...
+            self.context.siteObj = self.siteObj;
+
+            self.ts.registerLocal('site_root', function(flag, cb) {
+                cb(null, SiteService.getHostWithProtocol(self.siteObj.hostname) || self.ts.siteRoot);
+            });
+            self.ts.registerLocal('site_name', function(flag, cb) {
+                cb(null, self.siteName);
+            });
+            cb();
+        });
```
brianhyder: If the site is truly required for each request then I don't think it should be the responsibility of the controller to retrieve the information.  It should be handled as a step in the request pipeline (RequestHandler) before the controller is ever fired.

include/service/entities/site_service.js
```diff
@@ -0,0 +1,376 @@
+var async = require('async');
```
brianhyder: Would you please add our standard copyright to the top of the file?
```
/*
    Copyright (C) 2015  PencilBlue, LLC

    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.

    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
*/
```

include/service/entities/site_service.js
```diff
@@ -0,0 +1,376 @@
...
+
+    /**
+     * Service for performing site specific operations.
+     *
+     * @module Services
+     * @submodule Entities
+     * @class SiteService
+     * @constructor
+     */
+    function SiteService(){}
```
brianhyder: The site service is a huge part of the changes that are going in.  I'd like to get code level documentation on all of the properties and functions.

controllers/base_controller.js
```diff
@@ -116,54 +132,76 @@ module.exports = function BaseControllerModule(pb) {
...
             ts: this.ts,
-            activeTheme: this.activeTheme
+            site: this.site,
+            activeTheme: this.activeTheme,
+            onlyThisSite: true
         };
 
-        cb();
+        var siteService = new SiteService();
+        siteService.getByUid(this.site, function (err, siteInfo) {
```
brianhyder: It is extremely inefficient to have to fire off a DB call for EVERY request.  The site service should provide some kind of caching mechanism to avoid the extra DB call.  

There are a few ways to accomplish this: 
The easiest would be to use the simple layered service along with a configuration block to set the level of caching that can be used.  

At scale or under load this is really going hurt.  It is a simple optimization to save a shit ton of resources over time.

controllers/base_controller.js
```diff
@@ -186,6 +224,15 @@ module.exports = function BaseControllerModule(pb) {
     };
 
     /**
+     * @method getTemplateService
+     * @param {Object} options for TemplateService
+     * @return {Object} TemplateService
+     */
+    BaseController.prototype.getTemplateService = function(tsOps) {
+        return new pb.TemplateService(tsOps);
+    };
```
brianhyder: What was the need for this function?  Each controller already has an instance of TemplateService created during initialization.  Additional instances can be created by calling ```this.ts.getChildInstance()```.  The function will return a new instance and provide the same local registrations that the parent instance contained.

Was it just for mocking purposes?

include/admin_navigation.js
```diff
@@ -235,9 +271,9 @@ module.exports = function AdminNavigationModule(pb) {
      * @method getAdditions
      * @returns {Array}
      */
-    function getAdditions() {
-        return util.clone(AdminNavigation.additions);
-    }
+    function getAdditions(site) {
+        return getAdditionsInScope(AdminNavigation.additions, site);
+    };
```
brianhyder: only assigned functions need semi-colons

include/admin_navigation.js
```diff
@@ -257,14 +305,26 @@ module.exports = function AdminNavigationModule(pb) {
...
      */
-    function buildNavigation() {
+    function buildNavigation(site) {
         var i;
         var navigation = [];
-        var defaultNavigation = getDefaultNavigation();
-        var additions = getAdditions();
-        var childrenAdditions = getChildrenAdditions();
+        var multiSiteAdditions = getMultiSiteNavigation();
+        var defaultNavigation = getDefaultNavigation(site);
```
brianhyder: The calls for multi-stie navigation and default navigation don't always have to be made.  Only make the function call inside of the if/else block that they pertain too.

include/admin_navigation.js
```diff
@@ -379,46 +446,70 @@ module.exports = function AdminNavigationModule(pb) {
...
         return true;
     };
 
+    AdminNavigation.add = function (node) {
+        return AdminNavigation.addToSite(node, GLOBAL_SITE);
+    };
+
+    AdminNavigation.remove = function (id) {
+        return AdminNavigation.removeFromSite(id, GLOBAL_SITE);
+    };
```
brianhyder: There is no need for the ```add``` and ```remove``` functions.  The ```removeFromSite``` and ```addToSite``` should be modified to default the second parameter rather than wrapping in a default function.  It is a extra code to test and a performance degradation due to the extra function call.

include/admin_navigation.js
```diff
@@ -442,11 +533,12 @@ module.exports = function AdminNavigationModule(pb) {
...
-        AdminNavigation.additions = removeNode(id, AdminNavigation.additions);
+        AdminNavigation.additions[site] = removeNode(id, AdminNavigation.additions[site]);
 
-        util.forEach(AdminNavigation.childrenAdditions, function(node, parentId) {
-            AdminNavigation.childrenAdditions[parentId] = removeNode(id, node);
-        });
+        var childAdditionsMap = AdminNavigation.childrenAdditions[site];
+        for (var parentId in  childAdditionsMap) {
+            childAdditionsMap[parentId] = removeNode(id, childAdditionsMap[parentId]);
+        }
```
brianhyder: Code climate likes to complain about this kind of "for" loop:
```
The body of a for in should be wrapped in an if statement to filter unwanted properties from the prototype.
```

To solve this we created a utility function:
```
util.forEach(childAdditionsMap, function(value, key) {
  childAdditionsMap[key] = removeNode(id, value);
});
```

include/config.js
```diff
@@ -113,585 +93,287 @@ var CONFIG_MODULE_NAME  = 'config.js';
...
+
+        //the absolute file path to the directory where installation lives
+        docRoot:  Configuration.DOCUMENT_ROOT,
+
+        //enables/disables multiple sites in a single pencilblue instance (multitenancy)
+        multisite: false,
+
+        // When multisite is true, this is the hostname that will resolve to the global namespace.
+        // Only Admin routes will be activated for this hostname.
+        globalRoot: 'http://global.localhost:8080',
```
brianhyder: Multi-site is going to get more complex over time.  We need to add a block ```mutisite``` and it be an object rather than multiple top level properties.  This gives us a path forward rather than refactoring it at a later time with breaking changes.

include/config.js
```diff
@@ -113,585 +93,287 @@ var CONFIG_MODULE_NAME  = 'config.js';
...
-        authentication: {
-            un: null,
-            pw: null,
-            options: {
-                //authMechanism: "MONGODB-CR"|"GSSAPI"|"PLAIN", //Defaults to MONGODB-CR
-                //authdb: "db name here", //Defaults to the db attempted to be connected to
-                //authSource: "db name here", //Defaults to value of authdb
-            }
+Configuration.getBaseConfig = function(multisite) {
+    return util.clone({
```
brianhyder: If you are creating the object anyway there is no need to clone it.  I would prefer that we move it back out to its own variable ```BASE_CONFIG``` as before.  It gives more room for comments before the print margin and prevents the actual structure of the base configuration from being tampered with if the ```Configuration.getBaseConfig``` function is overriden.

include/config.js
```diff
@@ -751,9 +433,11 @@ Configuration.load = function(filePaths) {
  * @method mergeWithBase
  */
 Configuration.mergeWithBase = function(overrides) {
-    
+    var multisite = overrides ? overrides.multisite : false;
```
brianhyder: Why is this line necessary?

include/content.js
```diff
@@ -26,7 +26,10 @@ module.exports = function(pb) {
      * @class ContentService
      * @constructor
      */
-    function ContentService(){}
+    function ContentService(site, onlyThisSite) {
+        this.siteUid = pb.SiteService.getCurrentSite(site);
+        this.settingService = pb.SettingServiceFactory.getServiceBySite(this.siteUid, onlyThisSite);
+    }
```
brianhyder: For constructors, instead of multiple arguments we should pass an object.  This allows us to easily expand options without have to worry so much about breaking changes down the road.

include/dao/dao.js
```diff
@@ -85,6 +85,9 @@ module.exports = function DAOModule(pb) {
      */
     DAO.DESC = -1;
 
+    var GLOBAL_SITE = 'global';
```
brianhyder: This should probably be pulled from the configuration rather than a private variable in DAO.

include/dao/dao.js
```diff
@@ -710,6 +761,24 @@ module.exports = function DAOModule(pb) {
...
+                return cb(err);
+            }
+            db.collections(function(err, items) {
+                if (util.isError(err)) {
+                    return cb(err)
+                }
+                cb(err, items);
+            });
+        });
+    };
```
brianhyder: This function should be allowed to pass an options argument that contains a where clause to pass to the call to ```db.collections```.  This will allow ```DAO.entityExists``` to leverage the function

include/dao/dao.js
```diff
@@ -710,6 +761,24 @@ module.exports = function DAOModule(pb) {
...
     /**
+     * Gets all collection names
+     * @returns {DAO}
+     */
+    DAO.prototype.getAllCollections = function(cb) {
+        pb.dbm.getDb(this.dbName, function(err, db) {
+            if (util.isError(err)) {
+                return cb(err);
+            }
+            db.collections(function(err, items) {
```
brianhyder: The new mongo driver no longer supports calling ```collections```. 
http://mongodb.github.io/node-mongodb-native/2.0/api/Db.html#listCollections

Is this function really needed?

include/dao/dao.js
```diff
@@ -114,6 +117,48 @@ module.exports = function DAOModule(pb) {
         this.loadByValues(where, collection, opts, cb);
     };
 
+    DAO.prototype.loadByValueAvailableToSite = function(key, val, site, collection, opts, cb) {
```
brianhyder: Passing the site to every function seems very heavy handed.  The DAO constructor should except an options argument that is an object.  The object should not be required.  If object should support a ```site``` property that can be set for the instance of DAO.  This will prevent the extra parameters from being passed around.

include/dao/dao.js
```diff
@@ -114,6 +117,48 @@ module.exports = function DAOModule(pb) {
...
+            this.loadByValueFromGlobal(key,val,collection,opts,cb);
+        } else {
+            var where = {};
+            where[key] = val;
+            where[SITE_FIELD] = site;
+            this.loadByValues(where, collection, opts, cb);  
+        }
+    };
+
+    DAO.prototype.loadByValueFromGlobal = function(key, val, collection, opts, cb) {
```
brianhyder: All low level functions should be documented.  It is necessary for our understanding of code intentions as well as providing meaningful documentation in our auto-generated documentation.

include/dao/dao.js
```diff
@@ -413,6 +459,11 @@ module.exports = function DAOModule(pb) {
         });
     };
 
+    DAO.prototype.saveToSite = function(dbObj, site, options, cb) {
+        dbObj[SITE_FIELD] = site || GLOBAL_SITE;
+        this.save(dbObj, options, cb);
+    };
```
brianhyder: If the site context is set for the instance this function becomes unnecessary.

include/dao/db_manager.js
```diff
@@ -152,19 +152,33 @@ module.exports = function DBManagerModule(pb) {
...
+            var self = this;
             if (!util.isArray(procedures)) {
                 cb(new Error('The procedures parameter must be an array of Objects'));
                 return;
             }
 
+            this.dropUnconfiguredIndices(procedures, function(err) {
+                if(util.isError(err)) {
+                    pb.log.error('DBManager: Error occurred during index check/deletion ERROR[%s]', err.stack);
+                }
```
brianhyder: did you mean to silently fail here?

include/dao/db_manager.js
```diff
@@ -185,6 +199,91 @@ module.exports = function DBManagerModule(pb) {
...
+         * Sorts through all created indices and drops any index not declared in pb.config
+         * @method dropUnconfiguredIndices
+         * @param {Array} procedures
+         * @param {Function} cb
+         */
+        this.dropUnconfiguredIndices = function(procedures, cb) {
+            var self = this;
+            //to prevent a cirular dependency we do the require for DAO here.
+            var DAO = require('./dao.js')(pb);
+            this.getStoredIndices(function(err, storedIndices) {
```
brianhyder: The error is not handled

include/dao/db_manager.js
```diff
@@ -185,6 +199,91 @@ module.exports = function DBManagerModule(pb) {
...
+                var tasks = util.getTasks(storedIndices, function(indices, i) {
+                    return function(callback) {
+                        var index = indices[i];
+                        var filteredIndex = procedures.filter(function(procedure) {
+                            var ns = pb.config.db.name + '.' + procedure.collection;
+                            var result = ns === index.ns && self.compareIndices(index, procedure);
+                            return result;
+                        });
+                        var indexCollection = index.ns.split('.')[1];
+                        if(index.name !== '_id_' && indexCollection !== 'session' && (filteredIndex.length === 0 || util.isNullOrUndefined(filteredIndex))) {
```
brianhyder: The if statement is a bit complex.  Can you add some in-line comments as to what check are being done and why?

include/dao/db_manager.js
```diff
@@ -185,6 +199,91 @@ module.exports = function DBManagerModule(pb) {
...
+                            dao.dropIndex(indexCollection, index.name, function(err, result) {
+                                if(util.isError(err)) {
+                                    pb.log.error('DBManager: Failed to drop undeclared INDEX=[%s] RESULT=[%s] ERROR[%s]', JSON.stringify(index), util.inspect(result), err.stack);
+                                }
+                                callback(err, result);
+                            });
+                        }
+                        else {
+                            callback();
+                        }
```
brianhyder: Use the inverse of the if statement.  It will prevent the need to use the "else" statement and reduce nesting depth.

```
if(index.name === '_id_' || indexCollection === 'session' || (filteredIndex.length !== 0 && !util.isNullOrUndefined(filteredIndex))) {
  return callback();
}

//proceed to drop index here
```

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+    ];
+
+    var SITE_SPECIFIC_SETTINGS = [
+        'active_theme',
+        'content_settings',
+        'section_map',
+        'email_settings'
+    ];
+
+    function DBMigrate() {
```
brianhyder: Comments so it will show up in auto-docs

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+            siteService.getSiteMap(function (err, result) {
+                if (pb.config.multisite && result.active.length === 0 && result.inactive.length === 0) {
+                    self.createSite(function (err, isTaken, field, result) {
+                        self.siteUid = result.uid;
+                        var tasks = [
+                            util.wrapTask(self, self.migrateContentAndPluginData),
+                            util.wrapTask(self, self.migrateSettings),
+                            util.wrapTask(self, self.migrateUsers)
+                        ];
+                        //self.migrateContentAndPluginData(cb);
```
brianhyder: Is the commented code needed?

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+                        async.series(tasks, function(err, result) {
+                           cb(err, result);
+                        });
+                    });
+                }
+                else {
+                    cb(null, true);
+                }
```
brianhyder: Reverse the if statement to reduce nesting

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+            var site = pb.DocumentCreator.create('site', {
+                displayName: pb.config.siteName,
+                hostname: url.parse(pb.config.siteRoot).host
+            });
+            siteService.createSite(site, '', cb);
+        };
+
+        this.migrateContentAndPluginData = function(cb) {
+            var self = this;
+            var tasks = pb.util.getTasks(MIGRATE_ALL, function (collections, i) {
```
brianhyder: util is already required.  No need to reference it by the full namespace

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+                async.parallel(tasks, function(err, result) {
+                    cb(err, result);
+                });
+            });
+        };
+
+        this.migrateCollection = function (collection, siteUid, cb) {
+            var self = this;
+            var dao = new pb.DAO();
+            dao.q(collection, {}, function (err, results) {
```
brianhyder: The options parameter is optional.  It is not necessary to instantiate an empty object

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+                var tasks = util.getTasks(results, function(results, i) {
+                    return function(callback) {
+                      var uid = siteSpecificArr.indexOf(results[i][compareTo]) > -1? self.siteUid : pb.SiteService.GLOBAL_SITE;
+                      self.applySiteToDocument(results[i], uid, callback);
+                    };
+                });
+
+                async.parallel(tasks, function(err, result) {
+                    cb(err, result);
+                });
```
brianhyder: It is redundant to provide an anonymous function then use the callback.  Just pass the callback to ```parallel```
```
async.parallel(tasks, cb);
```

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+            var self = this;
+            var tasks = pb.util.getTasks(MIGRATE_ALL, function (collections, i) {
+                return function (callback) {
+                    self.migrateCollection(collections[i], self.siteUid, callback);
+                };
+            });
+
+            async.parallel(tasks, function (err, result) {
+                cb(err, result);
+            });
```
brianhyder: redundant anonymous function and callback

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+                        self.siteUid = result.uid;
+                        var tasks = [
+                            util.wrapTask(self, self.migrateContentAndPluginData),
+                            util.wrapTask(self, self.migrateSettings),
+                            util.wrapTask(self, self.migrateUsers)
+                        ];
+                        //self.migrateContentAndPluginData(cb);
+                        async.series(tasks, function(err, result) {
+                           cb(err, result);
+                        });
```
brianhyder: redundant anonymous function and callback

include/dao/db_migrate.js
```diff
@@ -0,0 +1,138 @@
...
+            dao.q(collection, {}, function (err, results) {
+                var tasks = util.getTasks(results, function (results, i) {
+                    return function (callback) {
+                        self.applySiteToDocument(results[i], siteUid, callback);
+                    };
+                });
+
+                async.parallel(tasks, function (err, result) {
+                    cb(err, result);
+                });
```
brianhyder: redundant anonymous function and callback

include/dao/indices.js
```diff
@@ -0,0 +1,327 @@
...
+ * Descending index value
+ * @private
+ * @static
+ * @readonly
+ * @property DESC
+ * @type {Integer}
+ */
+var DESC = -1;
+
+module.exports = function IndicesModule(multisite) {
```
brianhyder: This was a really good move to split this out.  Well done!

include/email.js
```diff
@@ -28,7 +28,11 @@ module.exports = function EmailServiceModule(pb) {
      * @class EmailService
      * @constructor
      */
-    function EmailService(){}
+    function EmailService(siteUid, onlyThisSite) {
+
+        this.site = pb.SiteService.getCurrentSite(siteUid);
+        this.onlyThisSite  = onlyThisSite;
+    }
```
brianhyder: Instead of parameterizing the constructor please use an object with properties.  Then error check and throw when required properties are not provided.  As irritating as it is to have things blow up I'd rather have the error message that tells me why I botched it.

include/email.js
```diff
@@ -61,7 +65,7 @@ module.exports = function EmailServiceModule(pb) {
      */
     EmailService.prototype.sendFromTemplate = function(options, cb){
         var self = this;
-        var ts   = new pb.TemplateService();
+        var ts   = new pb.TemplateService({ site: this.site });
```
brianhyder: I hate that I wrote it this way.  Need to add a "TODO" to move the instantiation of the template service to the constructor so it can be injectable with all of the other context properties it needs.

include/email.js
```diff
@@ -155,7 +159,8 @@ module.exports = function EmailServiceModule(pb) {
      */
     EmailService.prototype.getSettings = function(cb) {
         var self = this;
-        pb.settings.get('email_settings', function(err, settings) {
+        var settingsService = pb.SettingServiceFactory.getServiceBySite(self.site, self.onlyThisSite);
```
brianhyder: Same with the settings service.

include/http/request_handler.js
```diff
@@ -24,6 +24,7 @@ var async   = require('async');
 var domain  = require('domain');
 var Cookies = require('cookies');
 var util    = require('../util.js');
+var _ = require('lodash');
```
brianhyder: For my lack of familiarity with lodash, what does it by us here?

plugins/pencilblue/templates/admin/content/navigation/nav_map.html
```diff
@@ -35,7 +35,7 @@
         </div>
     </div>
     <div style="margin-top: 1em">
-        <a class="btn btn-default" href="/admin/content/navigation">
+        <a class="btn btn-default" ng-href="/admin/content/navigation">
```
blakecallens: Why the ng-href if it's not dynamic?

plugins/pencilblue/templates/admin/login.html
```diff
@@ -38,7 +38,7 @@
                                 </div>
                                 <input type="button" class="btn btn-lg btn-primary btn-block" value="^loc_LOGIN^" onclick="login()">
                                 <div class="forgot_password">
-                                    <a href="javascript:forgotPassword()">^loc_FORGOT_PASSWORD^</a>
+                                    <a href="javascript:forgotPassword()">I forgot my password</a>
```
blakecallens: 1. Why the change in wording?
2. Regardless, this should be localized.

plugins/pencilblue/templates/admin/plugins/plugin_settings.html
```diff
@@ -17,7 +17,7 @@
 						</div>
 					</div>
 				</div>
-				<a class="btn btn-default" href="/admin/plugins">
+				<a class="btn btn-default" ng-href="/admin/plugins">
```
blakecallens: No need for ng-href if it's not dynamic.

plugins/pencilblue/templates/admin/login.html
btidwell: This was Brian's change https://github.com/cbdr/pencilblue/commit/7a57e17c33ad0658dc25e874362dfade55b8cdc8

plugins/pencilblue/templates/admin/sites/site_form.html
```diff
@@ -0,0 +1,40 @@
...
+                        <input type="text" name="hostname" class="form-control" ng-disabled="isActive" ng-model="hostname" required>
+                        <div class="form_error" ng-if="!isActive && !isFieldValid(siteForm.hostname)">^loc_REQUIRED_FIELD^</div>
+                        <div class="form_error" ng-if="isActive">^loc_SITE_MUST_DEACTIVATE^</div>
+                    </div>
+                </div>
+                <br>
+                <a class="btn btn-default" href="/admin/sites">
+                    <i class="fa fa-ban"></i>&nbsp;^loc_CANCEL^
+                </a>
+                <a class="btn btn-primary" ng-click="newSite()" ng-if="isNew">
```
blakecallens: We can use the save button template and route to the correct functionality in the Angular controller.

plugins/pencilblue/templates/admin/sites/site_form.html
```diff
@@ -0,0 +1,40 @@
...
+                    <i class="fa fa-save"></i>&nbsp;^loc_SAVE^
+                </a>
+                <a class="btn btn-primary" ng-click="editSite()" ng-if="!isNew">
```
blakecallens: The save button template has a spinner for the async save call.

plugins/pencilblue/templates/angular/admin/plugins/manage_plugins.html
```diff
@@ -105,6 +104,17 @@
 			doJobRetrieve();
 		};
 
+		$scope.isInstalled = function(pluginName) {
+			var result = $.grep($scope.installedPlugins, function(currentPlugin){
```
blakecallens: Let's find a way to accomplish this with the jQuery lite in Angular, instead of injecting full jQuery into an Angular controller.

plugins/pencilblue/templates/angular/admin/sites/manage_sites.html
```diff
@@ -0,0 +1,134 @@
...
+    angular.module('pencilblueApp', ['validation'])
+            .controller('PencilBlueController', function($scope, $http, $window, validationService) {
+                ^angular_objects^
+
+                $scope.activateSite = function(site) {
+                    site.name = site.displayName;
+                    $scope.actionPlugin = site;
+                    $scope.actionProgress = '0';
+                    $scope.consoleOutput = '';
+                    $('#progress_modal').modal({});
```
blakecallens:
```JavaScript
angular.element('#progress_modal').modal();
```

plugins/pencilblue/templates/angular/admin/sites/manage_sites.html
```diff
@@ -0,0 +1,134 @@
...
+                                $scope.onActionFailure(error);
+                            });
+                };
+
+                $scope.deactivateSite = function(site) {
+                    site.name = site.displayName;
+                    $scope.actionPlugin = site;
+                    $scope.actionProgress = '0';
+                    $scope.consoleOutput = '';
+                    $('#progress_modal').modal({});
```
blakecallens: Same as above with angular.element

plugins/pencilblue/templates/angular/admin/sites/site_form.html
```diff
@@ -0,0 +1,74 @@
...
+
+                $scope.editSite = function() {
+                    $scope.formSubmitted = true;
+
+                    if(!validationService.isFormValid($scope.siteForm)) {
+                        return;
+                    }
+
+                    $scope.saving = true;
+                    console.log($scope.siteForm);
```
blakecallens: Looks like a debug log didn't get removed.

include/admin_navigation.js
```diff
@@ -56,115 +58,150 @@ module.exports = function AdminNavigationModule(pb) {
...
-                    href: '/admin/users/permissions',
-                    access: SecurityService.ACCESS_ADMINISTRATOR
-                },
-            ]
-        },
-        {
+    var MULTISITE_NAV = Object.freeze({
+            id: 'site_entity',
+            title: 'MANAGE_SITES',
+            icon: 'sitemap',
```
blakecallens: This icon makes sense for multisite, but that means we need to find a new one for custom objects.

include/admin_navigation.js
```diff
@@ -56,115 +58,150 @@ module.exports = function AdminNavigationModule(pb) {
...
-                    id: 'comments',
-                    title: 'COMMENTS',
-                    icon: 'comments',
-                    href: '/admin/content/comments',
-                    access: SecurityService.ACCESS_EDITOR
-                },
-                {
-                    id: 'custom_objects',
-                    title: 'CUSTOM_OBJECTS',
-                    icon: 'sitemap',
```
blakecallens: Suggest we change this to 'database', since multisite is now using 'sitemap'

plugins/pencilblue/templates/admin/login.html
blakecallens: So we can blame this on @brianhyder.

dissapointed

include/http/request_handler.js
```diff
@@ -606,6 +650,13 @@ module.exports = function RequestHandlerModule(pb) {
         //set the session
         this.session = session;
 
+        //set the site -- how do we handle improper sites here?
+        //TODO Handle global differently here when we pull through global site designation
+        this.siteObj = RequestHandler.sites[this.hostname]
+            ? RequestHandler.sites[this.hostname]
+            : this.serve404();
+        this.site = this.siteObj.uid;
+        this.siteName = this.siteObj.displayName;
```
brianhyder: Has the TODO already been handled?

include/localization.js
```diff
@@ -97,7 +92,7 @@ module.exports = function LocalizationModule(pb) {
      */
     Localization.prototype.localize = function(sets, text){
         if (pb.log.isSilly()) {
-            pb.log.silly('Localization: Localizing text - Locale [%s] Sets %s', this.language, JSON.stringify(sets));
+            pb.log.silly('Localization: Localizing text - Locale ['+this.language+'] Sets '+JSON.stringify(sets));
```
brianhyder: For readability please revert

include/localization.js
```diff
@@ -117,7 +112,7 @@ module.exports = function LocalizationModule(pb) {
 
         // If the localization is for HTML output, load the localization into client side JS
         if (text.indexOf('<body') > -1)  {
-            text = text.concat(pb.ClientJs.includeJS(pb.UrlService.createSystemUrl('api/localization/script?locale=' + this.language)));
+            text = text.concat(pb.ClientJs.includeJS(pb.config.siteRoot + '/localization/' + localizationLanguage + '.js'));
```
brianhyder: I think the merge might have gotten reversed on this file. These lines should be reversed.

include/localization.js
```diff
@@ -36,9 +33,7 @@ module.exports = function LocalizationModule(pb) {
      * @param {Object} request The request object
      */
     function Localization(request){
-        
-        //expected to be lowercase and of the form "en-us"
-        this.language = Localization.best(request).toString();
+        this.language = Localization.best(request).toString().toLowerCase().replace('-', '_');
```
brianhyder: I don't think the ```.toLowerCase``` and ```.replace``` are necessary anymore after the changes for allowing users to set their locale preference.

include/localization.js
```diff
@@ -141,7 +136,7 @@ module.exports = function LocalizationModule(pb) {
     Localization.prototype.get = function() {
         var key = arguments[0];
         if (pb.log.isSilly()) {
-            pb.log.silly('Localization: Localizing key [%s] - Locale [%s]', key, this.language);
+            pb.log.silly('Localization: Localizing key ['+key+'] - Locale ['+this.language+']');
```
brianhyder: this change was reversed

include/localization.js
```diff
@@ -163,8 +158,8 @@ module.exports = function LocalizationModule(pb) {
 
         if (val === null) {
             
-            var defaultLocale = Localization.getDefaultLocale().toLocaleLowerCase();
-            if (this.language === defaultLocale) {
+            var defaultLocale = Localization.getDefaultLocale();
+            if (this.language.toLowerCase() === defaultLocale.toLowerCase()) {
```
brianhyder: this change can be reversed

include/repository/plugin_repository.js
```diff
@@ -0,0 +1,204 @@
+var async = require('async');
+var util    = require('../util.js');
+
+module.exports = function PluginRepositoryModule(pb) {
+
+	var PLUGIN_COLL = 'plugin';
+	var GLOBAL_SITE = pb.SiteService.GLOBAL_SITE;
+	var SITE_FIELD = pb.SiteService.SITE_FIELD;
+
+	var publicAPI = {};
```
brianhyder: I'd like to get a little documentation on this file. Additionally, it differs a bit from the way the other files are structured.  I was curious if there was a specific reasoning behind that.

include/service/cache_entity_service.js
```diff
@@ -34,17 +34,23 @@ module.exports = function CacheEntityServiceModule(pb) {
...
+    function CacheEntityService(objType, valueField, keyField, site, onlyThisSite, timeout){
         this.type       = 'Cache';
         this.objType    = objType;
+        this.valueField = valueField ? valueField : null;
         this.keyField   = keyField;
+        this.site = site || GLOBAL_SITE;
+        this.onlyThisSite = onlyThisSite ? true : false;
         this.timeout    = timeout || 0;
-        this.valueField = valueField ? valueField : null;
     }
```
brianhyder: With the addition of the extra parameters we should convert this to a constructor that takes a single options object to prevent the parameters from getting out of control

include/service/cache_entity_service.js
```diff
@@ -55,36 +61,55 @@ module.exports = function CacheEntityServiceModule(pb) {
...
             if (util.isError(err)) {
                 return cb(err, null);
             }
 
-            //value doesn't exist in cache
+            //site specific value doesn't exist in cache
             if (result == null) {
-                return cb(null, null);
-            }
+                if(self.site !== GLOBAL_SITE && !self.onlyThisSite) {
```
brianhyder: Reverse the if logic and prevent the deep nesting.  
```
if (self.site === GLOBAL_SITE || self.onlyThisSite) {
  return cb(null, null)
}

File: include/service/db_entity_service.js [hunk 64]
```diff
@@ -31,13 +31,17 @@ module.exports = function DbEntityServiceModule(pb) {
...
      */
-    function DbEntityService(objType, valueField, keyField){
+    function DbEntityService(objType, valueField, keyField, site, onlyThisSite){
         this.type       = 'DB';
         this.objType    = objType;
         this.keyField   = keyField;
         this.valueField = valueField ? valueField : null;
+        this.site       = site || GLOBAL_SITE;
+        this.onlyThisSite       = onlyThisSite ? true : false;
     }
```
brianhyder: Probably time to change the constructor here too.

include/service/entities/plugin_service.js
```diff
@@ -36,7 +37,14 @@ module.exports = function PluginServiceModule(pb) {
      * @module Services
      * @submodule Entities
      */
-    function PluginService(){
+    function PluginService(siteUID){
```
brianhyder: The plugin service is going to have to change over to extend BaseObjectService at some point soon.  If we are going to change the constructor here I'd like to go ahead and make it consistent with the other services by passing an options object.

include/service/entities/plugin_service.js
```diff
@@ -161,7 +185,35 @@ module.exports = function PluginServiceModule(pb) {
...
+    PluginService.prototype.getAllActivePluginNames = function() {
+        var pluginNames = [];
+        var siteNames = Object.keys(ACTIVE_PLUGINS);
+        for( var i = 0; i < siteNames.length; i++ ) {
+            var sitePluginNames = Object.keys(ACTIVE_PLUGINS[siteNames[i]]);
+            for ( var j = 0; j < sitePluginNames.length; j++ ) {
+                pluginNames.push(siteNames[i] + '_' + sitePluginNames[j]);
+            }
+        }
+    };
```
brianhyder: Why not just return a hash where the key is the site name and the value is an array of active plugin UIDs.

include/service/entities/plugin_setting_service.js
```diff
@@ -0,0 +1,500 @@
...
+     * Null is provided in the event that the plugin is not installed.
+     */
+    PluginSettingService.prototype.getSettings = function(pluginName, cb) {
+        this.pluginSettingsService.get(pluginName, cb);
+    };
+
+    PluginSettingService.prototype.getSettingsBySite = function(pluginName, cb) {
+    	var settings = getAdminPluginSettingsService(this);
+    	settings.get(pluginName, cb);
+    }
```
brianhyder: semi-colon and documentation for function

include/service/entities/plugin_setting_service.js
```diff
@@ -0,0 +1,500 @@
...
+     * @param cb A callback that provides two parameters: cb(err, settingsObject)
+     */
+    PluginSettingService.prototype.getThemeSettings = function(pluginName, cb) {
+        this.themeSettingsService.get(pluginName, cb);
+    };
+
+    PluginSettingService.prototype.getThemeSettingsBySite = function(pluginName, cb) {
+    	var settings = getAdminThemeSettingsService(this);
+    	settings.get(pluginName, cb);
+    }
```
brianhyder: semi-colon and function documentation

include/service/entities/section_service.js
```diff
@@ -26,8 +26,14 @@ module.exports = function SectionServiceModule(pb) {
...
      * @constructor
+     * @param {String} site uid
+     * @param {Boolean} onlyThisSite should section service only return value set specifically by site rather than defaulting to global
      */
-    function SectionService(){}
+    function SectionService(site, onlyThisSite) {
+        this.site = pb.SiteService.getCurrentSite(site);
+        this.settings = pb.SettingServiceFactory.getServiceBySite(this.site, onlyThisSite);
+        this.siteQueryService = new pb.SiteQueryService(this.site, onlyThisSite);
+    }
```
brianhyder: Would like to move to options object parameter

include/service/entities/site_query_service.js
```diff
@@ -0,0 +1,208 @@
...
+    DAO.prototype.save.call(this, dbObj, options, callback);
+  };
+
+  /**
+   * Gets all collection names
+   * @param {Function} cb
+   */
+  SiteQueryService.prototype.getCollections = function (cb) {
+    var dao = new pb.DAO();
+    dao.listCollections({}, function(err, items) {
```
brianhyder: if SiteQueryService inherits from DAO why do you need a new instance here to call ```listCollections```?  Can you not just do:
```
SiteQueryService.prototype.getCollections = function (cb) {
  this.listCollections({}, cb);
};
```

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+                 self.getActiveSites(callback);
+             },
+
+             inactive: function(callback) {
+                 self.getInactiveSites(callback);
+             }
+        };
+        async.series(tasks, function(err, results) {
+            cb(err, results);
+        });
```
brianhyder: unnecessary anonymous callback.  Just use the ```cb``` parameter

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+        dao.q(SITE_COLL, {select: pb.DAO.SELECT_ALL, where: {uid: uid} }, function(err, result) {
+            var siteName = (!uid || uid === SiteService.GLOBAL_SITE) ? 'global' : '';
+
+            if(pb.util.isError(err)) {
+                pb.log.error(err);
+            }else if(result && result.length > 0) {
+                siteName = result[0].displayName;
+            }
+
+            cb(siteName);
```
brianhyder: We try to standardize on callback structure.  It should always pass any error as the first parameter or null if does not exist.

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+     * @param {String} uid - unique id
+     * @param {Function} cb - the callback function
+     */
+    SiteService.prototype.getSiteNameByUid = function(uid, cb) {
+        var dao = new pb.DAO();
+        dao.q(SITE_COLL, {select: pb.DAO.SELECT_ALL, where: {uid: uid} }, function(err, result) {
+            var siteName = (!uid || uid === SiteService.GLOBAL_SITE) ? 'global' : '';
+
+            if(pb.util.isError(err)) {
+                pb.log.error(err);
```
brianhyder: Why do we silently fail here instead of reporting the error back through the callback?

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+        var getWhere = function(where) {
+            if (id) {
+                where[pb.DAO.getIdField()] = pb.DAO.getNotIDField(id);
+            }
+            return where;
+        };
+        var dao   = new pb.DAO();
+        var tasks = {
+            displayName: function(callback) {
+                var expStr = util.escapeRegExp(displayName.toLowerCase) + '$';
```
brianhyder: You are passing a function rather than passing a display name.  I think it should be:
```
var expStr = '^' + util.escapeRegExp(displayName.toLowerCase()) + '$';
```

brianhyder: Side question do you want the '^' character at the beginning to force start of line matching.

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+                dao.count('site', getWhere({displayName: new RegExp(expStr, 'i')}), callback);
+            },
+            hostname: function(callback) {
+                dao.count('site', getWhere({hostname: hostname.toLowerCase()}), callback);
+            }
+        };
+        async.series(tasks, cb);
```
brianhyder: why run these in series vs. parallel?

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+            });
+        }
+    };
+
+    // Generate a site unique id.
+    function getUid() {
+        return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+            var r = Math.random()*16|0,v=c=='x'?r:r&0x3|0x8;return v.toString(16);
+        });
+    }
```
brianhyder: Well I'm curious. How does this work?

brianhyder: Is this a specific format you all needed or would util.uniqueId not work for you?

include/service/entities/site_service.js
```diff
@@ -0,0 +1,491 @@
...
+    SiteService.siteExists = function(siteUid, cb) {
+        if (pb.config.multisite.enabled && !(siteUid === SiteService.GLOBAL_SITE)) {
+            var dao = new pb.DAO();
+            dao.exists(SITE_COLL, {uid: siteUid}, function (err, exists) {
+                cb(err, exists);
+            });
+        }
+        else {
+            cb(null, true);
+        }
```
brianhyder: reverse the if logic and avoid the nesting

include/service/entities/url_service.js
```diff
@@ -31,7 +31,11 @@ module.exports = function UrlServiceModule(pb) {
      * @module Services
      * @submodule Entities
      */
-    function UrlService() {}
+    function UrlService(site, onlyThisSite) {
+        this.site = pb.SiteService.getCurrentSite(site);
+        this.onlyThisSite = onlyThisSite;
+        this.siteQueryService = new pb.SiteQueryService(this.site, this.onlyThisSite);
+    }
```
brianhyder: I like to use an options object for constructor parameters

include/service/entities/user_service.js
```diff
@@ -242,29 +252,33 @@ module.exports = function(pb) {
...
-        var emailService = new pb.EmailService();
-        emailService.getSettings(function(err, emailSettings) {
-            var options = {
-                to: user.email,
-                replacements: {
-                    verification_url: pb.UrlService.createSystemUrl('/actions/user/verify_email?email=' + encodeURIComponent(user.email) + '&code=' + encodeURIComponent(user.verification_code)),
-                    first_name: user.first_name,
-                    last_name: user.last_name
+        var siteService = new pb.SiteService();
+        siteService.getByUid(self.context.site, function(err, siteInfo) {
```
brianhyder: Need to do error checking

include/service/entities/user_service.js
```diff
@@ -242,29 +252,33 @@ module.exports = function(pb) {
...
+            // We need to see if email settings have been saved with verification content
+            var emailService = new pb.EmailService(self.context.site);
+            emailService.getSettings(function (err, emailSettings) {
```
brianhyder: need to do error checking

include/service/entities/user_service.js
```diff
@@ -277,23 +291,27 @@ module.exports = function(pb) {
...
-            replacements: {
-                'verification_url': verficationUrl,
-                'first_name': user.first_name,
-                'last_name': user.last_name
-            }
-        };
-        var emailService = new pb.EmailService();
-        emailService.sendFromTemplate(options, cb);
+        var siteService = new pb.SiteService();
+        siteService.getByUid(self.context.site, function(err, siteInfo) {
```
brianhyder: need to perform error checking

include/service/jobs/sites/site_activate_job.js
```diff
@@ -0,0 +1,105 @@
+
+//dependencies
+var async = require('async');
+var util  = require('../../../util.js');
+
+module.exports = function SiteActivateJobModule(pb) {
+    var GLOBAL_PREFIX = 'global';
```
brianhyder: get this value from the SiteService so the value is only hard coded in one place

include/service/jobs/sites/site_activate_job.js
```diff
@@ -0,0 +1,105 @@
...
+        var siteUid   = this.getSite();
+        var tasks     = [
+            //set site to active in mongo
+            function(callback) {
+                var dao = new pb.DAO();
+                dao.loadByValue('uid', siteUid, 'site', function(err, site) {
+                    if(util.isError(err)) {
+                        callback(err, null)
+                    } else if (!site) {
+                        callback(new Error('Site not found'), null);
```
brianhyder: "return" the callbacks on error conditions and avoid having to declare an else block.  Will save you from having to nest the else conditions so deep.

include/service/jobs/sites/site_activate_job.js
```diff
@@ -0,0 +1,105 @@
...
+                });
+            }
+        ];
+        async.series(tasks, function(err, results) {
+            if(util.isError(err)) {
+                cb(err);
+                return;
+            }
+            cb(null, true);
+        });
```
brianhyder: you can take it down to a one liner:
```
cb(err, !util.isError(err));
```

include/service/jobs/sites/site_deactivate_job.js
```diff
@@ -0,0 +1,102 @@
...
+        var siteUid   = this.getSite();
+        var tasks     = [
+            //set site to active in mongo
+            function(callback) {
+                var dao = new pb.DAO();
+                dao.loadByValue('uid', siteUid, 'site', function(err, site) {
+                    if(util.isError(err)) {
+                        callback(err, null)
+                    } else if (!site) {
+                        callback(new Error('Site not found'), null);
```
brianhyder: "return" callbacks and avoid else block

include/service/jobs/sites/site_deactivate_job.js
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                });
+            }
+        ];
+        async.series(tasks, function(err, results) {
+            if(util.isError(err)) {
+                cb(err);
+                return;
+            }
+            cb(null, true);
```
brianhyder: one liner the callback ```cb(err, !util.isError(err));```

include/service/jobs/sites/site_deactivate_job.js
```diff
@@ -0,0 +1,102 @@
+
+//dependencies
+var async = require('async');
+var util  = require('../../../util.js');
+
+module.exports = function SiteDeactivateJobModule(pb) {
+    var GLOBAL_PREFIX = 'global';
+
+    function SiteDeactivateJob(){
```
brianhyder: Documentation for prototype so it shows up in auto-docs

include/service/jobs/sites/site_activate_job.js
```diff
@@ -0,0 +1,105 @@
+
+//dependencies
+var async = require('async');
+var util  = require('../../../util.js');
+
+module.exports = function SiteActivateJobModule(pb) {
+    var GLOBAL_PREFIX = 'global';
+
+    function SiteActivateJob(){
```
brianhyder: Documentation for prototype so it shows up in auto-docs

include/service/jobs/sites/site_job_runner.js
```diff
@@ -0,0 +1,58 @@
+var util = require('../../../util.js');
+
+module.exports = function SiteJobRunnerModule(pb) {
+
+    function SiteJobRunner() {
```
brianhyder: Documentation for prototype so it shows up in auto-docs

include/theme/top_menu.js
```diff
@@ -92,10 +95,17 @@ module.exports = function TopMenuServiceModule(pb) {
      * @method getAccountButtons
      * @param {Object}   session
      * @param {Object}   ls      The localization service
+     * @param {String}   site    The current site
```
brianhyder: Need to make it ```@param {String} [site] The current site```
The square brackets tell the documentation parser that the parameter is optional

include/util.js
```diff
@@ -394,6 +394,18 @@ Util.copyArray = function(array) {
...
+    var a = array.concat();
+    for(var i = 0; i < a.length; ++i) {
+        for(var j=i+1; j<a.length; ++j) {
+            if(a[i] === a[j]) {
+                a.splice(j--, 1);
+            }
+        }
+    }
+    return a;
+};
```
brianhyder: I think you will get the same performance if not better and less code with:
```
Util.dedupeArray = function(array) {
    var hash = Util.arrayToHash(array);
    return Object.keys(hash);
};
```

pencilblue.js
```diff
@@ -128,6 +130,21 @@ function PencilBlue(config){
...
     };
 
+    this.initSiteMigration = function(cb) {
+        pb.dbm.processMigration(cb);
+    };
+
+    /**
+     *
+     */
+    this.initSites = function(cb)
```
brianhyder: documentation for the two new tasks in the startup sequence.

plugins/pencilblue/controllers/actions/admin/sites/activate_site.js
```diff
@@ -0,0 +1,45 @@
...
+        if(message) {
+            self.formError(message, '/admin/sites', cb);
+            return;
+        }
+
+        var siteService = new pb.SiteService();
+        var jobId = siteService.activateSite(vars.id);
+        var content = pb.BaseController.apiResponse(pb.BaseController.API_SUCCESS, '', jobId);
+        cb({content: content});
+    };
```
brianhyder: I think this controller mixes two concepts.  The error checking calls upon a function that deals with forms but the happy path calls back with an API response.  

Is this a view controller or API controller?

plugins/pencilblue/controllers/actions/admin/sites/deactivate_site.js
```diff
@@ -0,0 +1,45 @@
...
+        var message = this.hasRequiredParams(vars, ['id']);
+        if(message) {
+            self.formError(message, '/admin/sites', cb);
+            return;
+        }
+
+        var siteService = new pb.SiteService();
+        var jobId = siteService.deactivateSite(vars.id);
+        var content = pb.BaseController.apiResponse(pb.BaseController.API_SUCCESS, '', jobId);
+        cb({content: content});
```
brianhyder: Same thing here. mix of form and Api constructs

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
```diff
@@ -0,0 +1,50 @@
...
+
+    DeleteSiteAction.prototype.render = function(cb) {
+        var self = this;
+        var siteid = self.pathVars.siteid;
+        var siteQueryService = new pb.SiteQueryService();
+        siteQueryService.getCollections(function(err, allCollections) {
+            siteQueryService.deleteSiteSpecificContent(allCollections, siteid, function(err, result) {
+                if(util.isError(err)) {
+                    cb({
+                        code: 400,
```
brianhyder: Is it a bad request (400) or a server error (500)?  PB allows you to callback with an error that has the message and ```code``` property set.  It will handle the formatting and sending of the specified code:

Some service:
```
Service.prototype.doSomething(cb) {
 var failures = BaseObjectService.validationFailure('someField', 'someField is required');
 var err = BaseObjectService.validationError([failures]);
 cb(err);
};
```

Controller (inside of handler):
```
this.service.doSomething(function(err, result) {
  if (util.isError(err)) {
    return cb(err);
  }
});
```

Server Output: HTTP Status - 400
```
{
  "message": "Validation Error",
  "validationErrors": [
    {
      "field": "someField",
      "message": "someField is required"
    }
  ]
}
```

plugins/pencilblue/controllers/actions/admin/sites/edit_site.js
```diff
@@ -0,0 +1,89 @@
...
+     * Edits a current site
+     */
+    function EditSiteAction(){}
+    util.inherits(EditSiteAction, pb.BaseController);
+
+    EditSiteAction.prototype.render = function(cb)
+    {
+        var self = this;
+        var siteid = this.pathVars.siteid;
+        this.getJSONPostParams(function(err, post) {
```
brianhyder: Instead of calling this, in the route declaration include ```request_body: ['application/json']```.  It will instruct the request handler to automatically parse the body for you.  You can specify multiple mime types in the array block.  In your controller you no longer need to call ```getJSONPostParams```.  You can just refer to the parsed body as: ```this.body```.  

It removes the extra level of nesting and extra function call by the controller.

plugins/pencilblue/controllers/actions/admin/sites/edit_site.js
```diff
@@ -0,0 +1,89 @@
...
+                });
+                return;
+            }
+
+            var siteService = new pb.SiteService();
+            var dao = new pb.DAO();
+            dao.loadByValue('uid', siteid, 'site', function(err, data) {
+                siteService.isDisplayNameOrHostnameTaken(post.displayName, post.hostname, data._id, function (err, isTaken, field) {
+                    if(isTaken) {
+                        cb({
```
brianhyder: "return" the callback and remove the else block to reduce nesting

plugins/pencilblue/controllers/actions/admin/sites/new_site.js
```diff
@@ -0,0 +1,82 @@
...
+     * Creates a new site
+     */
+    function NewSiteAction(){}
+    util.inherits(NewSiteAction, pb.BaseController);
+
+    NewSiteAction.prototype.render = function(cb)
+    {
+        var self = this;
+
+        this.getJSONPostParams(function(err, post) {
```
brianhyder: specify request_body mime types in route definition to avoid having to make this call manually.

plugins/pencilblue/controllers/actions/admin/sites/new_site.js
```diff
@@ -0,0 +1,82 @@
...
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+module.exports = function NewSiteActionModule(pb) {
```
brianhyder: Route handler functions can be specified as part of the route definition.  If you specify the handler then you can group common actions such as "edit site" and "new site" in the same file and reduce the amount of code that needs to be written and maintained.

plugins/pencilblue/templates/angular/admin/sites/manage_sites.html
IanDTrapp: Fixed, I'll push it up with my other changes

plugins/pencilblue/templates/angular/admin/sites/manage_sites.html
IanDTrapp: Fixed, I'll push it up with my other changes

plugins/pencilblue/templates/angular/admin/sites/site_form.html
IanDTrapp: Fixed, I'll push it up with my other changes

include/admin_navigation.js
cchatham: @blakecallens do we need to add this or are y'all taking care of this on your side?

.gitignore
btidwell: I have removed the duplicate entry

controllers/admin/base_admin_controller.js
btidwell: I have updated this with your suggestion.

controllers/base_controller.js
btidwell: This has been removed.

controllers/base_controller.js
btidwell: Removed reference to pb.util.

controllers/base_controller.js
btidwell: This is modified with the change of the DB lookup in base controller.  Let me know what you think of the implementation.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

controllers/base_controller.js
btidwell: Added all site info required into the site_map of request handler.  The DB request has been removed.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

controllers/base_controller.js
btidwell: This function has been removed.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/admin_navigation.js
btidwell: Done.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/admin_navigation.js
btidwell: These have been moved accordingly.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/admin_navigation.js
btidwell: This has been modified.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/dao.js
btidwell: This is fixed.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_manager.js
btidwell: No I did not.  Now the error is passed into a callback with a return to end the function.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_manager.js
btidwell: Same change made here.  Used a callback with  return instead.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_manager.js
btidwell: Done.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_manager.js
btidwell: This is done as well with a comment explaining what is being evaluated.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_migrate.js
btidwell: No longer needed.  It has been removed.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_migrate.js
btidwell: Done.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_migrate.js
btidwell: Done.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_migrate.js
btidwell: Done.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/dao/db_migrate.js
btidwell: Done.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/http/request_handler.js
btidwell: Here is a link to it.  https://lodash.com/  It is mainly used for formatting collections such as removing duplicate values as we do in the request_handler on line 780.

"themesToCheck = _.uniq(themesToCheck);"

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/http/request_handler.js
btidwell: Yes removing it now.

Ben Tidwell
Software Engineer III | Talent Network
CareerBuilder
5550-A Peachtree Pkwy, Norcross GA 30092

include/admin_navigation.js
blakecallens: @cchatham if you can switch the custom object one to fa-cubes in this PR, it'd be much appreciated.
http://fontawesome.io/icon/cubes/

include/service/entities/user_service.js
wesanderson303: error handling added to getByUid().

include/service/entities/user_service.js
wesanderson303: error handling added to getSettings().

include/service/entities/user_service.js
wesanderson303: Error handling added to sendPasswordResetEmail()

include/dao/dao.js
cchatham: @brianhyder is there anyway we can break out this refactor out of this PR? I would like to a bigger refactor that would touch more files and this PR is already crazy. It would be great if we could iterate.

include/email.js
cchatham: @brianhyder is there an action item here for us? Or just notes for you guys?

include/localization.js
cchatham: @brianhyder Do we really want to make that assumption? If yes, then we just need to revert this line?

include/email.js
brianhyder: For us.  I won't pawn refactoring our poor code off onto you guys.  

However, would you do me a small favor and add the TODO in there with my comments?

include/dao/dao.js
brianhyder: That is perfectly fine.  As long as we have a game plan I'm happy.

include/service/jobs/sites/site_activate_job.js
cchatham: @brianhyder I agree that the callbacks need to be returned, but in this case why would we remove the else? Is it javascript bad practice to complete if statements with an else?

include/service/jobs/sites/site_deactivate_job.js
wesanderson303: Consolidated

include/service/jobs/sites/site_activate_job.js
brianhyder: No, not bad practice at all.  Just a different pattern.  All it really saves you the need the extra indention if you use the return statements.  
```
if(util.isError(err)) {
	return callback(err, null)
}
if (!site) {
	return callback(new Error('Site not found'), null);
} 
site.active = true;
dao.save(site, function(err, result) {
	if(util.isError(err)) {
		return cb(err, null);
	}

	pb.RequestHandler.activateSite(site);
	callback(err, result)
});
```

include/localization.js
brianhyder: Yes, I believe the merge on this file was incorrect.  The changes in this file are a reversal of the changes I made for of commit: https://github.com/pencilblue/pencilblue/commit/f7bcb6fee3190332343c2aaaa6dd97cf63e3191c

I think it is safe to make the assumption.  Any bugs that result will be from the change mentioned above and we will address as a bug fix.

include/service/entities/site_service.js
wesanderson303: Included callback with error

include/service/entities/site_service.js
wesanderson303: First param is now error results (null in this case). Updated handler as well for this

plugins/pencilblue/templates/angular/admin/sites/manage_sites.html
btidwell: Modal is now on angular instead of jquery.

plugins/pencilblue/templates/angular/admin/sites/manage_sites.html
btidwell: Done in the same commit for angular.element

plugins/pencilblue/templates/angular/admin/sites/site_form.html
btidwell: log has been removed.

include/admin_navigation.js
btidwell: @blakecallens I changed the icon for custom objects to cubes.

include/admin_navigation.js
btidwell: This was changed with what was suggested above.

plugins/pencilblue/templates/admin/sites/site_form.html
btidwell: Swapped out links for buttons with icon animations.

plugins/pencilblue/templates/admin/sites/site_form.html
btidwell: This has been changed to save button as well.

plugins/pencilblue/controllers/actions/admin/sites/new_site.js
andparker: I believe we opted for separate files for "edit site" and "new site" to maintain consistency with how the other admin actions (content, users) were constructed.

brianhyder: We are trying to move away from that model because it creates a lot of
duplicate code. It is a nice to have rather than a must have.

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
andparker: Hey Brian. It looks like only BaseObjectService objects right now are using the validation methods. Do you want to keep this the convention or do you not mind if other service types use these functions?

brianhyder: The goal is to eventually get everything consistent with the way BaseObjectService handles things since that will be the majority of controllers will operate (controller calls service for heavy lifting).

brianhyder:
Thanks y'all!!!!!!!!!!!!!!!!!!!

blakecallens:
Woot!

PR #684, opened 2015-07-12
Localization changes: fix for bug #667

kylebechtel77:
Active themes localization folder takes precedence over latest installed plugin now. If the localized variable doesnt exist in the active themes localization folder it will search the other plugins localization files and use the first one it finds.

### DIFF
--- a/include/http/request_handler.js
+++ b/include/http/request_handler.js
@@ -405,9 +405,6 @@ module.exports = function RequestHandlerModule(pb) {
      */
     RequestHandler.prototype.handleRequest = function(){
 
-        //get locale preference
-        this.localizationService = new pb.Localization(this.req);
-
         //fist things first check for public resource
         if (RequestHandler.isPublicRoute(this.url.pathname)) {
             this.servePublicContent();
@@ -603,6 +600,8 @@ module.exports = function RequestHandlerModule(pb) {
                 pb.log.warn("RequestHandler: The active theme is not set.  Defaulting to '%s'", RequestHandler.DEFAULT_THEME);
                 activeTheme = RequestHandler.DEFAULT_THEME;
             }
+            //get locale preference
+            self.localizationService = new pb.Localization(self.req, activeTheme);
             self.onThemeRetrieved(activeTheme, route);
         });
     };
--- a/include/http/request_handler.js
+++ b/include/http/request_handler.js
@@ -405,6 +405,9 @@ module.exports = function RequestHandlerModule(pb) {
      */
     RequestHandler.prototype.handleRequest = function(){
 
+        //get locale preference
+        this.localizationService = new pb.Localization(this.req);
+
         //fist things first check for public resource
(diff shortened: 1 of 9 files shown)


brianhyder:
@kylebechtel77 thank you for submitting a fix. I'll be reviewing the changes over the next day or so. Could you please update the unit tests to fix the 2 that are failing? 

Thanks again for submitting ting a fix.

kylebechtel77:
Yes, Ill look at the unit tests over the next couple of days and push the fixes to this branch when I get them done. Thanks for responding so quickly

brianhyder:
One other request please.  We are preparing to release 0.4.1 in the next week and I'd like to include these changes in that release.  Would you mind closing out this PR and creating another that would merge into our ```0.4.1``` branch?  It would be greatly appreciated.  

We recently added the ability to set the user's locale and in doing so have made some changes to the request_handler.js file as well as the localization.js file.  I'd like to work with you to get those merged into 0.4.1.

include/http/request_handler.js
```diff
@@ -603,6 +600,8 @@ module.exports = function RequestHandlerModule(pb) {
                 pb.log.warn("RequestHandler: The active theme is not set.  Defaulting to '%s'", RequestHandler.DEFAULT_THEME);
                 activeTheme = RequestHandler.DEFAULT_THEME;
             }
+            //get locale preference
+            self.localizationService = new pb.Localization(self.req, activeTheme);
```
brianhyder: You will have to initialize the localization service sooner.  The implementation here will cause an error when a non-existent route is provided.  See line ~590.  

I think what you need is a setter on the Localization service that allows you to set the prioritized theme.  By default, or when provided null or undefined it would default to ```pencilblue```.

include/localization.js
```diff
@@ -35,8 +35,8 @@ module.exports = function LocalizationModule(pb) {
      * @constructor
      * @param {Object} request The request object
      */
-    function Localization(request){
-        
+    function Localization(request, activeTheme){
+        this.activeTheme = activeTheme;
```
brianhyder: If the setter implementation is used you will not need the extra parameter on the constructor.  However, I would go ahead and set the default value on the property to ```pencilblue```

include/service/entities/plugin_service.js
```diff
@@ -1216,7 +1216,7 @@ module.exports = function PluginServiceModule(pb) {
                              pb.log.debug('PluginService:[%s] Registering localizations for locale [%s]', details.uid, locale);
                          }
 
-                         var result = pb.Localization.registerLocalizations(locale, localizations[locale]);
+                         var result = pb.Localization.registerLocalizations(locale, localizations[locale], plugin.dirName);
```
brianhyder: Please use ```details.uid``` here.  There are instances where the folder name does not match the plugin UID.  We are trying to shy away from that and deprecate that functionality.

kylebechtel77:
closing pull request to open another on branch 0.4.1 with the changes Brian specified.

PR #689, opened 2015-07-18
fix permissions on admin settings

klavs:
None of these `hrefs` are actually available to anyone except admin as defined in `plugins/pencilblue/include/routes.js`


brianhyder:
Thank you for fix!

Issue #677, opened 2015-07-06
Latest installed plugins localization file takes precedence over active theme 


kylebechtel77:
There are several plugins that have some localized fields that have the same key e.g. 404_error (located in their localization/en-us.json file). I would expect pencilblue to use the active themes localization files, or use the localization files from the plugin that the page's route belongs to, but it does not. Instead, it takes the latest one it comes to in the list of installed plugins.

To replicate this I cloned down the latest pb project, installed sample and set it to active and viewed its page on sample/random/text. Then I edited portfolios localization/en-us.json file to contain "SAMPLE_HELLO_WORLD": "Hello World!!! from portfolio", and installed it, set it to the active theme and viewed it to make sure there were no issues. I then set sample to the active theme again and viewed its page and it said Hello World!!! from portfolio.

brianhyder:
@kylebechtel77 This should have been fixed in the latest merge into 0.5.0.  If you run into issues again please let us know.

PR #696, opened 2015-07-22
updates to go into 0.5.0 to fix bug #677

kylebechtel77:
fixed conditionals. tests now pass

### DIFF
--- a/include/localization.js
+++ b/include/localization.js
@@ -303,13 +303,13 @@ module.exports = function LocalizationModule(pb) {
      * @param {Object} localizations
      * @return {Boolean}
      */
-    Localization.registerLocalizations = function(locale, localizations) {
+    Localization.registerLocalizations = function(locale, localizations, pluginName) {
         if (!Localization.isSupported(locale) || !util.isObject(localizations)) {
             return false;
         }
         
         util.forEach(localizations, function(item, key) {
-            Localization.registerLocalization(locale, key, item);
+            Localization.registerLocalization(locale, key, item, pluginName);
         });
         return true;
     };
--- a/include/localization.js
+++ b/include/localization.js
@@ -36,10 +36,14 @@ module.exports = function LocalizationModule(pb) {
      * @param {Object} request The request object
      */
     function Localization(request){
-        
-        //expected to be lowercase and of the form "en-us"
-        this.language = Localization.best(request).toString();
+      this.activeTheme = 'pencilblue';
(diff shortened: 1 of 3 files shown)


kylebechtel77:
Do not pull this in yet, just found an issue. working to resolve it now.

kylebechtel77:
Issue is resolved. this can be pulled in now.

include/localization.js
```diff
@@ -153,7 +157,14 @@ module.exports = function LocalizationModule(pb) {
         var tmp;
         var val = null;
         var loc = Localization.storage[this.language];
-        for (var category in loc) {
+
+        if(loc && loc[this.activeTheme] && loc[this.activeTheme][key])
+        {
+          return loc[this.activeTheme][key];
+        }
```
brianhyder: Item 1

include/localization.js
```diff
@@ -312,31 +323,27 @@ module.exports = function LocalizationModule(pb) {
...
+        pluginName = 'generic';
+
+      var localeObj = Localization.getLocalizationPackage(locale);
+
+      if (localeObj[pluginName] !== undefined) {
+        localeObj[pluginName][key] = value;
+      }
+      else {
+        localeObj[pluginName] = {};
+        localeObj[pluginName][key] = value;
```
brianhyder: Item 2

brianhyder:
Hey all, I wanted to pass along some feedback on this PR.  We won't be merging this in but I wanted to take a minute to explain why.  

1. There is a bug at "Item 1".  The code find the key and returns the value if there is a plugin override.  This breaks current functionality where parameterized values are not replaced when overrided by a plugin.
2. The data structure used to store plugin overrides has the potential to cause collisions with top level keys in the localization files.
3. There are API end-points that provide the localizations client side.  These would need to be updated too in order to provide backward compatibility as well as the ability to apply plugin overrides through some query string parameter.
4. There are already flaws in our current architecture with localization.  Specifically, the way we fail to distinguish between language by itself and when a country code is paired with a language code.  Fixing this will require a different data structure that splits out language, country, and themes.  There will also be additional complexity around multi-site and how that is handled.

Part of the work I am targeting for this quarter revolves around fixing these flaws and building language capabilities into the routes.  I really want to get the multi-site merged in before this work begins to avoid nasty merges and to create a base-line.  

I'd like to use this code as the starting point for that work.  The overall feel is that the fix for the bug is a part of a larger issue with the way things are structured.  

Let me know if y'all have any questions.

PR #701, opened 2015-07-28
Enable easy override of template paths

markbaird:
This change to ContentViewLoader allows controllers to easily override template paths without the need to override entire render functions.


brianhyder:
Looks Good!

PR #711, opened 2015-08-12
breaks edit and create site into separate functions

kylebechtel77:


cchatham:
We found a bug with edit and new site. This fixes it.

PR #714, opened 2015-08-13
Serves an error for an invalid hostname

andparker:
Serves an error in the request handler when the request host is not registered or is not the siteroot.


brianhyder:
Awesome. Thanks @andparker!!!

PR #715, opened 2015-08-18
Site create edit job runner

btidwell:
When we update/create sites, the information is not replicated across all servers in the request handler.  As a result Sites are not loaded until the app servers restart.


include/service/entities/site_service.js
```diff
@@ -276,29 +297,18 @@ module.exports = function SiteServiceModule(pb) {
...
-                    return;
-                }
-
-                cb(null, false, null, result);
-            });
-        });
+    SiteService.prototype.createSite = function(options, cb) {
+        cb = cb || util.cb;
+        options.active = false;
+        options.uid = pb.util.uniqueId();
```
brianhyder: ```util``` is already declared.  ```pb.util``` is redundant.

brianhyder:
Look good @btidwell!  I'm going to give @blakecallens a chance to look over the front end piece and then I'm good to merge.

blakecallens:
Looks good to me!

PR #728, opened 2015-08-24
Feature/model registration

brianhyder:


plugins/sample/controllers/random_text_view.js
```diff
@@ -157,7 +174,11 @@ module.exports = function RandomTextViewControllerModule(pb) {
                 //as the second argument to specify that the value should not be 
                 //HTML encoded.
                 cb(null, new pb.TemplateValue(content, false));
-            });
+            };
+            
+            //here we register the model so that our values will be rendered in 
+            //the view
+            self.ts.registerModel(model/*, optional model name can be provided here */);
```
brianhyder: @blakecallens FYI, this is the model (all puns intended) that we want to move forward with.

Issue #739, opened 2015-09-08
Issue using new events emitted by BaseObjectService


markbaird:
I'm working on some plugins to integrate with third party tools and services (CloudFlare and Sendy). I need to be able to make some API calls to those third party services when an article is created or updated. So I created a new plugin and added the following line to the main plugin module:
`BaseObjectService.on('article.' + BaseObjectService.AFTER_SAVE, CloudFlareService.afterSave);`

However my `afterSave()` function is never called. I tried tracking down the issue by setting breakpoints in BaseObjectService, but they were never hit. So I viewed the actual API calls being made by the browser when the admin saves a page, and those are going to `ArticleApiController -> BaseAdminController -> SiteQueryService -> DAO`. So it looks like the ArticleService/BaseObjectService isn't being used at all when an article is saved.

Am I just going about this all wrong? Or is this a known issue, currently being worked on? It seems that the SiteQueryService is a wrapper for the DAO (it says so in the comments) that is bypassing the object service implementations. It seems like it would be better if it called the object services (ArticleService, PageService, etc.) but it isn't clear to me how it would know which one to call.

brianhyder:
@markbaird can you provide the URL that you are hitting to expect your afterSave function to be called?

My guess is that it is the end-point called when the article is saved from the admin UI.  If this is the case then you are correct.  The afterSave function would not be called.  We haven't yet gotten all of the controllers converted over to utilize the new article service.  

Not 100% sure if it is as simple as changing the resource URL in the form or if the controller needs to be updated.  Most likely, it is both.  

@blakecallens was targeting the conversion of the controllers between now and the 4th quarter.

markbaird:
Yes it is the end-point that is called when the article is saved from the admin UI. So basically I'm trying to use a feature that isn't implemented yet.

I'd be interested in helping to implement this feature, given some guidance on how you envisioned it working. It doesn't appear to be as simple as swapping out the ArticleService for the ArticleServiceV2 in the controller.

Thanks

brianhyder:
We would be thrilled to have the help.

UI:
* Change the URI path to the new API end-point: ```/api/content/articles```
* Ensure the date fields send the values back in ISO format as a string
* Ensure the meta-keywords are sent back as an array of strings
* Update portfolio form to do the same thing

Controllers (new_article.js & edit_article.js):
* Deprecate the controller in the prototype comments

markbaird:
I'll work on this. You can assign the issue to me if you want.

[...]

PR #741, opened 2015-09-09
#739 Admin API end-point change

markbaird:
Partial changes for #739 

This converts all article operations in the Admin UI to use the new API end-point. If these changes are acceptable I will make similar changes to all page operations.

### DIFF
--- a/include/service/entities/content/article_service_v2.js
+++ b/include/service/entities/content/article_service_v2.js
@@ -140,12 +140,6 @@ module.exports = function(pb) {
         dto.url = BaseObjectService.sanitize(dto.url);
         dto.publish_date = BaseObjectService.getDate(dto.publish_date);
         
-        if (util.isArray(dto.meta_keywords)) {
-            for (var i = 0; i < dto.meta_keywords.length; i++) {
-                dto.meta_keywords[i] = BaseObjectService.getDate(dto.meta_keywords[i]);  
-            }
-        }
-        
         cb(null);
     };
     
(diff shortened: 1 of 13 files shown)


include/service/entities/content/article_service_v2.js
```diff
@@ -140,12 +140,6 @@ module.exports = function(pb) {
         dto.url = BaseObjectService.sanitize(dto.url);
         dto.publish_date = BaseObjectService.getDate(dto.publish_date);
         
-        if (util.isArray(dto.meta_keywords)) {
-            for (var i = 0; i < dto.meta_keywords.length; i++) {
-                dto.meta_keywords[i] = BaseObjectService.getDate(dto.meta_keywords[i]);
```
markbaird: This was attempting to parse meta_keyword values as dates, which was obviously causing problems.

markbaird: It just occurred to me that perhaps this section should remain, but call sanitize() instead of getDate()? Was that the original intention?

brianhyder: Yup, that was it.  I botched that.  Good catch!

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
```diff
@@ -130,6 +132,26 @@
...
+							articleData.meta_keywords.push(keywords[i]);
+						}
+					}
+				}
+			}
+
+			// This sometimes defaults to "undefined" instead of "false" on page load
+			if (typeof articleData.allow_comments == "undefined") {
+				articleData.allow_comments = false;
+			}
```
brianhyder: I would just simplify this to: ```articleData.allow_comments = articleData.allow_comments ? true : false;```

brianhyder:
Just that one small change and I'm content.

markbaird:
This resolves #739

This pull request now includes all the changes to the Admin UI (and Portfolio theme) to use the new article and page API end-points.

PR #742, opened 2015-09-10
Stop customObjectService.fetchChildren call if custom object type is not found.

veimis:
Giving an invalid custom object id to customObjectService.fetchChildren results in a fields-property being read from a null object, thus throwing an error. 
```
var tasks = util.getTasks(Object.keys(custObjType.fields), function(fieldNames, i) {
```
where custObjType is null.

I was giving the ID of the custom object as a string instead of the ID of the custom object type for the fetchChildren().

### DIFF
--- a/include/service/entities/custom_object_service.js
+++ b/include/service/entities/custom_object_service.js
@@ -358,9 +358,10 @@ module.exports = function CustomObjectServiceModule(pb) {
 
         //make sure we have the type for the object passed in
         getCustObjType(custObjType, function(err, custObjType) {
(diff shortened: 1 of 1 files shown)


veimis:
Failing to provide a valid custom object type could be considered an error and the fix would be something like this
```
else if(util.isNullOrUndefined(custoObjType)) {
    return cb(new Error('Custom object type not found.'));
}
```

brianhyder:
Awesome! Thank you for submitting the fix.

PR #752, opened 2015-09-19
Allow configuration of media root URL

markbaird:
This change allows you to configure the root URL to be used for media uploaded by PencilBlue. This allows you to use a different subdomain like cdn.mydomain.com for media served by PencilBlue, which is often necessary for CDN caching of static resources. You can also use this in conjunction with a media provider like the s3-pencilblue plugin to send all media request traffic directly to your media storage server.

You don't need to re-upload media when you change the media root URL.

This change also displays the configured media root URL on the site settings page.

### DIFF
--- a/include/config.js
+++ b/include/config.js
@@ -354,6 +354,11 @@ Configuration.getBaseConfig = function(multisite) {
             provider: 'fs',
             parent_dir: 'public',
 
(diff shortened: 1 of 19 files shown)


brianhyder:
Looks good @markbaird. Thanks!

PR #753, opened 2015-09-20
Fix for subnav pills failing to load on media list page

markbaird:
I copied the logic from manage_articles.js to get this working. The previous code was trying to set the `pills` attribute on the variable `angularObjects` which is a String, resulting in the nav buttons for adding new media never displaying.


blakecallens:
Thanks for the fix. I noticed this while working in 0.5.0, so there's a fix inside of the factories bramch too.

PR #762, opened 2015-09-24
Media loader site awareness

evanadams28:
When committing multi-tenancy, we overlooked the media loader. This PR is site-awareness for the media loader. This is required for a multi-tenant media loader. Please review. Thanks!


include/service/entities/content/article_renderer.js
```diff
@@ -1,42 +1,44 @@
...
      */
-    function ArticleRenderer(context) {
+    function ArticleRenderer(context, options) {
         if (context) {
             this.commentService = new pb.CommentService(context);
             this.hostname = context.hostname;
         }
+        this.site = pb.SiteService.getCurrentSite(options.site);
+        this.onlyThisSite = options.onlyThisSite;
     }
```
brianhyder: There isn't a need for an options object here.  Just provide the property as part of the context object.  They are synonymous.

include/service/entities/content/article_service_v2.js
```diff
@@ -34,18 +34,19 @@ module.exports = function(pb) {
      * @constructor
      * @extends BaseObjectService
      */
-    function ArticleServiceV2(context){
+    function ArticleServiceV2(context, options){
```
brianhyder: There is no need for an options object.  Use the context parameter to provide site which is probably already in the context object.

include/service/entities/custom_object_service.js
```diff
@@ -361,9 +361,10 @@ module.exports = function CustomObjectServiceModule(pb) {
 
         //make sure we have the type for the object passed in
         getCustObjType(custObjType, function(err, custObjType) {
-            if (util.isError(err)) {
+            if (util.isError(err) || util.isNullOrUndefined(custObjType)) {
```
brianhyder: Calling back with only an empty error would be incorrect behavior here.  There should be a condition such that:
```
else if (util.isNullOrUndefined(custObjType)) {
  return cb(new Error('An invalid custom object type: ' + custObjType + ' was found.'));
}
```

plugins/pencilblue/controllers/section.js
```diff
@@ -36,10 +36,11 @@ module.exports = function(pb) {
      */
     SectionViewController.prototype.init = function(context, cb) {
         var self = this;
+        var opts = { site: self.site, onlyThisSite: self.onlyThisSite };
         var init = function(err) {
             //get content settings
             var serviceContext = self.getServiceContext();
-            var contentService = new pb.ContentService({site: self.site, onlyThisSite: serviceContext.onlyThisSite});
+            var contentService = new pb.ContentService(opts);
```
brianhyder: check the service context to see if it contains the two options you want to pass.

include/service/entities/content/article_renderer.js
evanadams28: Oh wow, not sure how I didn't catch that. I'll go ahead and remove options everywhere there is already a context.

evanadams28:
I've made your recommended changes. :)

include/service/entities/content/article_service_v2.js
```diff
@@ -38,14 +38,15 @@ module.exports = function(pb) {
         if (!util.isObject(context)) {
             context = {};
         }
-        
+        this.site = this.site = pb.SiteService.getCurrentSite(context.site);
```
brianhyder: Why the redundant assignment?

include/service/entities/content/article_service_v2.js
```diff
@@ -38,14 +38,15 @@ module.exports = function(pb) {
...
+        this.onlyThisSite = context.onlyThisSite;
```
brianhyder: for the new properties can you please add the documentation blocks?

```
/**
 *
 * @property [name here]
 * @type {Boolean}
 */
```

plugins/pencilblue/controllers/section.js
```diff
@@ -39,32 +39,32 @@ module.exports = function(pb) {
         var init = function(err) {
             //get content settings
             var serviceContext = self.getServiceContext();
-            var contentService = new pb.ContentService({site: self.site, onlyThisSite: serviceContext.onlyThisSite});
+            var contentService = new pb.ContentService(context);
```
brianhyder: I think you want to pass the service context.  The reason you call the ```getServiceContext``` is because it creates a separate object to pass to each service.  This prevents the individual services from modifying the context object and causing issues across other services.

plugins/pencilblue/controllers/section.js
```diff
@@ -39,32 +39,32 @@ module.exports = function(pb) {
...
             contentService.getSettings(function(err, contentSettings) {
                 if (util.isError(err)) {
                     return cb(err);
                 }
                 //create the service
                 self.contentSettings = contentSettings;
-                var asContext = self.getServiceContext();
+                var asContext = serviceContext;
```
brianhyder: I would call get ```self.getServiceContext()``` here as it was for the reason noted above.

brianhyder:
@evanadams28 Just a couple of small changes and I'll click merge. I promise!

Really good job on this too.

include/service/entities/content/article_service_v2.js
evanadams28: honest typo

include/service/entities/content/article_service_v2.js
evanadams28: absolutely!

plugins/pencilblue/controllers/section.js
evanadams28: Ah, very nice separation of control. I like it! Will change.

plugins/pencilblue/controllers/section.js
evanadams28: `asContext` was actually doing that already and I changed it to this. `serviceContext` was an unused variable prior, so I took advantage of it rather than deleting it. However, the change above will be making use of it again so I'll change this back for the same reasoning as above.

evanadams28:
@brianhyder Changes done! Thanks for the feedback.

blakecallens:
@evanadams28, congratulations on surviving a @brianhyder code review!

Magnifying Glass

brianhyder:
:+1:

evanadams28:
Whoo! Thanks @brianhyder! Deleting branch.

PR #763, opened 2015-09-24
SiteObjectService for site event hooks

wesanderson303:
We created a SiteObjectService inheriting from BaseObjectService to handle the BEFORE_SAVE and AFTER_DELETE event hooks so we could integrate that process with AWS.


include/service/entities/site_object_service.js
```diff
@@ -0,0 +1,134 @@
...
+   * @extends BaseObjectService
+   */
+  function SiteObjectService(context) {
+    if (!util.isObject(context)) {
+      context = {};
+    }
+
+    context.type = TYPE;
+    SiteObjectService.super_.call(this, context);
+    this.siteService = new SiteService();
```
brianhyder: Why didn't you just make the current SiteService extend from BaseObjectService?  Now there is an additional prototype that does almost the same thing and makes it more confusing as to which one should be used.

include/service/entities/site_object_service.js
```diff
@@ -0,0 +1,134 @@
...
+
+    pb.BaseObjectService.on(TYPE + '.' + pb.BaseObjectService.MERGE, SiteObjectService.merge);
```
brianhyder: You don't want to register events here.  Placing the registration here will cause the event to be registered and fired for each instance of the service that is created.  This will cause a memory leak because for each instance you will accrue an additional reference pointer to the merge function

brianhyder: You want to move the registrations to the bottom of the file.  Use the TopicService as an example.

include/service/entities/site_object_service.js
```diff
@@ -0,0 +1,134 @@
...
+   * A service interface function designed to allow developers to name the handle
+   * to the service object what ever they desire. The function must return a
+   * valid string and must not conflict with the names of other services for the
+   * plugin that the service is associated with.
+   *
+   * @static
+   * @method getName
+   * @return {String} The service name
+   */
+  SiteObjectService.getName = function () {
```
brianhyder: This function is not necessary for core services.  It is only used for plugin specific services.

include/service/entities/site_object_service.js
```diff
@@ -0,0 +1,134 @@
...
+    return SERVICE_NAME;
+  };
+
+  SiteObjectService.merge = function (context, cb) {
+    pb.util.merge(context.data, context.object);
+    cb(null);
+  };
+
+  SiteObjectService.prototype.deleteSiteSpecificContent = function (collections, siteid, callback) {
```
brianhyder: You can simplify this whole process.  The BaseObjectService already provides a "delete" mechanism.  Use that and tap into the hooks (```BaseObjectService.BEFORE_DELETE``` or ```BaseObjectService.AFTER_DELETE```)to delete the site specific data.

include/service/entities/site_object_service.js
```diff
@@ -0,0 +1,134 @@
...
+            taskCallback(null, {collection: collections[i].name, recordsDeleted: numberOfDeletedRecords});
+          });
+        };
+      });
+      async.parallel(tasks, function (err, results) {
+        if (pb.util.isError(err)) {
+          pb.log.error(err);
+          return callback(err);
+        }
+        self.dao.delete({uid: siteid}, 'site', function (err/*, result*/) {
```
brianhyder: The BaseObjectService already provides a generic "delete" implementation, as mentioned above.  Using that will allow other developers to perform actions of their own when sites are being removed.

include/service/entities/site_object_service.js
wesanderson303: I need to delete a variety of site-related fields. Is there a way I can hook into a batch delete event instead of each one? The way I'm handling it here, I send the AFTER_DELETE event after the entire delete batch is completed.

include/service/entities/site_object_service.js
wesanderson303: In reference to my above note, read below. I emit the AFTER_DELETE manually instead of hooking into BOS's deleteSingle() event hooks.

include/service/base_object_service.js
```diff
@@ -581,14 +581,14 @@ module.exports = function(pb) {
                 if (util.isError(err)) {
                     return cb(err, null);
                 }
-                
-                self.dao.delete(options.where, self.type, options, function(err, result) {
+
+                self.dao.delete(options.where, options.collection, options, function(err, result) {
```
wesanderson303: Reference site_service line 586 below for why this is needed.

include/service/base_object_service.js
```diff
@@ -581,14 +581,14 @@ module.exports = function(pb) {
...
                     if (util.isError(err)) {
-                        return cb(err, obj);
+                        return cb(err, result);
```
wesanderson303: Reference site_service line 593 below for why this is needed

include/service/entities/site_service.js
```diff
@@ -554,5 +569,54 @@ module.exports = function SiteServiceModule(pb) {
...
+                return callback(err);
+            }
+
+            if (!exists) {
+                return callback(pb.BaseObjectService.validationError([pb.BaseObjectService.validationFailure("siteid", "Invalid siteid")]));
+            }
+
+            var tasks = util.getTasks(collections, function (collections, i) {
+                return function (taskCallback) {
+                    self.deleteSingle({where: {site: siteid}, collection: collections[i].name}, function (err, commandResult) {
```
brianhyder: Using delete single here doesn't make sense.  Using a specific object service type here for other object types can be dangerous.  It will emit events for objects that it doesn't apply to.  In this particular case just use the DAO prototype so that you can manually specify the collections and not alter the behavior of the BaseObjectService.  

The long term solution will be to create a factory such that BaseObjectService instances for a specific type can be looked up through a single function.  Until then it will be fine to delete the site specific data for each collection through DAO.

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
```diff
@@ -30,12 +30,13 @@ module.exports = function DeleteSiteActionModule(pb) {
         var self = this;
         var siteid = self.pathVars.siteid;
         var siteQueryService = new pb.SiteQueryService();
+        var SiteService = new pb.SiteService();
```
brianhyder: Please use variables that start with a lowercase character for instance variable names.  We typically reserver initial upper camel case for prototype names.

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
```diff
@@ -30,12 +30,13 @@ module.exports = function DeleteSiteActionModule(pb) {
...
         siteQueryService.getCollections(function(err, allCollections) {
-            siteQueryService.deleteSiteSpecificContent(allCollections, siteid, function(err, result) {
-                if(util.isError(err)) {
+            SiteService.deleteSiteSpecificContent(allCollections, siteid, function(err, result) {
```
brianhyder: Why not call the code shown below instead?

```
siteService.deleteSingle({ where: { uid: siteid}}, function(err, site) {
  ...
});
```
The above implies that the SiteService registers a before or after delete event to remove site specific content.  

* Is it really important to get the number of records deleted?

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
wesanderson303: oops, typo. Good catch!

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
wesanderson303: To your previous comment though, won't this leave leftover site content? We still need to delete all content for a site, right?

include/service/entities/site_service.js
wesanderson303: Good call. So we can just call dao.delete() here and siteService.deleteSingle below for the last event hook?

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
brianhyder: Yes, it would.  To solve this problem, inside of the SiteService, I would register a function for the AFTER_DELETE that would:
* retrieve the collection list
* iterate over collections deleting documents that contain the given site id.

SiteService.js
```
SiteService.prototype.deleteSiteSpecificContent = function(siteUid, options, cb) {

  //retrieve list of collections
  //for each collection
    //delete from collection[i] where siteId = 'xyz'
 //callback when done
};

SiteService.afterDelete = function(context, cb) {
  var siteObj = context.data;
  context.service.deleteSiteSpecificContent(siteObj.uid, cb);
};

//Event Registries
BaseObjectService.on(TYPE + '.' + BaseObjectService.AFTER_DELETE, SiteService.afterDelete);
```

This is basically a reversal of the logic in the implementation you have for ```deleteSiteSpecificContent```.  The approach I am suggesting will simplify the amount of code you have to write.  You will not need to verify the site exists because the delete single check does that for you and you don't worry about manually having to delete the site object after all other site data was deleted.  All you will need in the outlined ```deleteSiteSpecificContent``` function is what is on 584-597 of site_service.js and the code to retrieve the collections list.

wesanderson303: Ah. I like it. I'll play with this approach and update the PR ASAP. Thanks Brian!

include/service/jobs/sites/site_create_edit_job.js
```diff
@@ -100,7 +101,7 @@ module.exports = function SiteCreateEditJobModule(pb) {
                     site.hostname = mySite.hostname || site.hostname;
                     site.displayName = mySite.displayName || site.displayName;
 
-                    dao.save(site, function(err, result) {
+                    siteService.save(site, function(err, result) {
```
brianhyder: Nice!!!

plugins/pencilblue/controllers/actions/admin/sites/new_site.js
```diff
@@ -52,12 +52,13 @@ module.exports = function NewSiteActionModule(pb) {
             if(isTaken) {
                 return cb({
                     code: 400,
-                    content: pb.BaseController.apiResponse(pb.BaseController.API_ERROR, self.ls.get('DUPLICATE_INFO'))
+                    content: pb.BaseController.apiResponse(pb.BaseController.API_FAILURE, self.ls.get('DUPLICATE_INFO'))
```
brianhyder: Good catches!!!

brianhyder:
Looks good! Thanks @wesanderson303!

PR #764, opened 2015-09-25
400 controller override and 500 page

kylebechtel77:
This allows for the 400 controller to be over-written on a plugin by plugin basis.  This allows for plugins to do custom things in their serve 404 pages.  Previously they only were allowed to override the html file for the 404 but that was all. 

In the spirit of this change, we also added the ability to make a 500 page (like what is currently possible with the 404 page.)  While we were at it, we went ahead and added in the ability to override the controller for this 500 page just like the change we added for the 500 page.  

Let me know if you have any questions,

Logan Genet & Kyle Bechtel


include/http/request_handler.js
```diff
@@ -538,57 +538,74 @@ module.exports = function RequestHandlerModule(pb) {
...
+      } catch(e){
+        NotFound = require(path.join(pb.config.docRoot, 'plugins/pencilblue/controllers/error/404.js'))(pb);
+      }
+
+      var cInstance = new NotFound();
+      self.doRender({}, cInstance, {});
+
+      if (pb.log.isSilly()) {
+        pb.log.silly("RequestHandler: No Route Found, Sending 404 for URL=" + this.url.href);
+      }
```
brianhyder: Neat trick to save you a couple of lines and keep performance.  The logging functions provide the ability to pass in parameters to the log statement such they can be formatted in.  Adding parameters to the function stack doesn't cost much.  About as much as doing the if statement. 
```
pb.log.silly('RequestHandler: No Route Found, Sending 404 for URL=%s', this.url.href);
```
The above will only do the replacement if the log level is "silly".

kylebechtel77: We have done that. Thanks for the tip.

lgenet:
Hey, are there any updates on this?  Anything we need to change so that you all will merge it?

brianhyder:
We'll try and get around to it this week.  We are addressing a few other items at this time.

brianhyder:
Hey guys,  my apologies on the extended delay.  I was struggling with the breaking changes that this would cause with the ErrorFormatter prototype.  I didn't want to reduce functionality by providing a solution that would have to be duplicated for each error code.  This is my fault as I said that would be fine in our email exchange about design.  I have another pull request opened: #790.  It takes the changes that you all have proposed and included them in the Error formatters.  

This will allow additional overrides at a theme, code, and default level.  I've added comments to the PR at the important bits.  If you get a chance take a look and see if I overlooked anything.  

Also wanted to say thanks for laying the ground work on this one.  It made it really easy to copy over the changes from this PR and integrate them into the error formatters.

Let me know if y'all have any questions.

PR #772, opened 2015-09-28
This is a fix for site query service

kylebechtel77:
This uses base node to remove the system libraries rather than relying on the mongod plugin. We did this to fix potential deployment problems with aws elastic beanstalk that others might run into.


include/service/entities/site_query_service.js
```diff
@@ -260,13 +260,20 @@ module.exports = function SiteQueryServiceModule(pb) {
...
    */
   SiteQueryService.prototype.getCollections = function (cb) {
-    this.listCollections({ name: { $nin: ["system.indexes", "system.namespaces"] }}, function(err, items) {
-      if(pb.util.isError(err)) {
-        pb.log.error(err);
-      }
-      cb(err, items);
-    });
-  };
+      this.listCollections({}, function(err, items) {
```
brianhyder: It will probably be more efficient to hand off the filtering to the DB.  The object you pass as the first parameter is the filter object.  You could specify that the name should not match a regular expression and filter it that way rather than writing additional code.

include/service/entities/site_query_service.js
```diff
@@ -260,13 +260,20 @@ module.exports = function SiteQueryServiceModule(pb) {
...
+          if(pb.util.isError(err)) {
+            pb.log.error(err);
+          }
+
+          for(var i = 0; i<items.length; i++) {
+              if(items[i].name.indexOf('system.indexes') != -1 || items[i].name.indexOf('system.namespaces') != -1){
+                items.splice(i, 1);
+              }
+          }
```
brianhyder: In the event that you do need to filter them out at the web-server use the [Array.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) function.  It will save a couple of lines.

```
items = items.filter(function(item) {
  return item.name.indexOf('system.indexes') === -1 && item.name.indexOf('system.namespaces') === -1;
});
```

include/service/entities/site_query_service.js
kylebechtel77: The issue we ran into with handing the filtering off to the db was that it was dependent on several things, from your version of node to mongod npm package you had installed. taking a slightly less efficient route that guaranteed the proper results seemed like the best way to solve the problem, and it did so we figured we would push it back to you so others dont run into the same issue.

include/service/entities/site_query_service.js
kylebechtel77: Will switch to items.filter. Thanks for the recommendation.

include/service/entities/site_query_service.js
brianhyder: Fair enough.  Thank you for the clarifications.  Much appreciated.

include/service/entities/site_query_service.js
```diff
@@ -260,13 +260,18 @@ module.exports = function SiteQueryServiceModule(pb) {
...
-      }
-      cb(err, items);
-    });
-  };
+      this.listCollections({}, function(err, items) {
+          if(pb.util.isError(err)) {
+            pb.log.error(err);
+          }
+
+          items = items.filter(function(item) {
```
brianhyder: One more small change and I am good to merge.  If there is an error it will be logged.  However, the ```items``` parameter will most likely be null or undefined.  In the ```if``` statement above we should callback with the error and return otherwise we may run into a null reference error here on line 268 when we attempt to filter the array.

kylebechtel77: Excellent suggestion. I should have caught that. Ill take care of it shortly.

brianhyder:
@kylebechtel77 There was a merge conflict after merging another PR.  Would you mind updating and I'll merge everything in.

kylebechtel77:
I believe I have fixed the merge conflicts. Let me know if there are any more issues.

PR #787, opened 2015-10-16
Updates to support media file uploads that result in remote absolute url locations

joshisa:
Great project!  I've been building a PencilBlue Media Provider Plugin for the Openstack Swift Bluemix Object Storage v1 service.   After trying a few approaches, I realized with the current code base that a file upload didn't seem able to return a remote HTTP url location.  My media provider takes the content stream from the file upload and "puts" the content into the remote system. This results in a mediaPath value that is an absolute HTTP URL.  This is sort of a hybrid of the Media URL Link usecase and the local file upload use case.  

Without this PR, my provider was always embedding a media entry that was treated as a relative file on the server rather than a remote linked absolute url resource.  Please advise if there is a better way to approach the issue.  I've tried to maintain backwards compatibility by falling back to the default behavior if the new remoteURL argument is a boolean.

### DIFF
--- a/include/service/entities/media_service.js
+++ b/include/service/entities/media_service.js
@@ -238,8 +238,12 @@ module.exports = function MediaServiceModule(pb) {
      */
     MediaService.prototype.setContent = function(fileDataStrOrBuff, fileName, cb) {
         var mediaPath = MediaService.generateMediaPath(fileName);
-        this.provider.set(fileDataStrOrBuff, mediaPath, function(err, result) {
-            cb(err, { mediaPath: mediaPath, result: result});
+        this.provider.set(fileDataStrOrBuff, mediaPath, function(err, remoteURL, result) {
+            if (util.isBoolean(remoteURL)) {		
+              cb(err, { mediaPath: mediaPath, result: result});		
+            } else {		
+              cb(err, { mediaPath: remoteURL, result: result});		
+            }
         });
     };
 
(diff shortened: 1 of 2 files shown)


brianhyder:
@joshisa Just wanted to give you a heads up.  I haven't had a chance to look over the PR yet.  Been a little swamped.  Hoping to get to it this week.  Thank you for the contribution!

joshisa:
@brianhyder Cool.  No worries.  Maybe we'll run into each other.  I'm at All Things Open ... will check out the pencilblue prez tomorrow.

include/service/entities/media_service.js
```diff
@@ -238,8 +238,12 @@ module.exports = function MediaServiceModule(pb) {
      */
     MediaService.prototype.setContent = function(fileDataStrOrBuff, fileName, cb) {
         var mediaPath = MediaService.generateMediaPath(fileName);
-        this.provider.set(fileDataStrOrBuff, mediaPath, function(err, result) {
-            cb(err, { mediaPath: mediaPath, result: result});
+        this.provider.set(fileDataStrOrBuff, mediaPath, function(err, remoteURL, result) {
+            if (util.isBoolean(remoteURL)) {		
+              cb(err, { mediaPath: mediaPath, result: result});		
+            } else {		
+              cb(err, { mediaPath: remoteURL, result: result});
```
brianhyder: Can we make the ```remoteURL``` a property of ```result```?  This prevents the parameter bleed.

plugins/pencilblue/controllers/actions/admin/content/media/delete_media.js
```diff
@@ -77,7 +77,7 @@ module.exports = function(pb) {
     };
 
     DeleteMediaController.prototype.removeLocal = function(media, mservice, cb) {
-        if (!media.is_file) {
+        if (!media.is_file && media.location.indexOf('http') !== 0 && media.location.indexOf('//') !== 0) {
```
brianhyder: If you do this then how does the file actually get deleted out of the storage provider?

My guess is that there is no way for the providers to set a remote URL and a storage path.

joshisa: The provider's implemented delete method would receive the absolute url contained in the media.location.  The provider then issues an HTTP DELETE against the absolute url provided to remove it.  Alternatively, might there be a way to facilitate the provider to toggle the is_file property boolean to their liking?  This would then eliminate the need to check for an absolute url.

include/service/entities/media_service.js
joshisa: Agreed.  Avoiding parameter bleed would be good.  I'll make some mods to test that approach.

blakecallens:
Closing this because it's obsolete.

PR #791, opened 2015-10-21
Adding missing semicolon to HTML entity

markbaird:
The `&nbsp` HTML entity in the read more link was missing the semicolon. This was causing rendering issues in system generated emails.


brianhyder:
Thanks @markbaird!

PR #794, opened 2015-10-28
Dutch translation

jacobsyentl:
Dutch translation for PencilBlue.


blakecallens:
Thanks for doing this! Is the nl-BE standard for "belgian dutch", or should
it be nl-NL?

jacobsyentl:
Well nl-NL is practically the same as nl-BE. I wrote it like that because I live in Belgium, but it doesn't matter. Both countries speak the same language :)

blakecallens:
If that's the case, can you duplicate the file and also submit a nl-NL.js?
Thanks!

brianhyder:
Please do not duplicate the file. Pb is smart enough to house that locale
based  on the language alone.

brianhyder:
*pick

blakecallens:
My only concern is if, over time, we might want to add Flemish (true nl-BE) separately.

brianhyder:
At that point all you will need is to provide the specific differences. All
other keys will default to the other locale.

blakecallens:
Well, OK then. :)

jacobsyentl:
So I don't need to add an extra file then?

blakecallens:
No, I was mistaken.

blakecallens:
We good to merge this, @brianhyder?

PR #801, opened 2015-11-05
Update Polish translation

wvffle:
Some of the keys were missing and translation was kind of bad in some strings.
There were some untranslated values as well.


brianhyder:
Awesome, thank you!

wvffle:
No problem :)

PR #804, opened 2015-11-06
updates to dao and site query service

kylebechtel77:
This is just some cleanup we've done to the dao and sqs files to clean up some unneeded code. Dao was still using references to site and we removed those and put them in sqs instead. Since this is a change to base pencilblue we wanted to push it back to you.


brianhyder:
Thanks @kylebechtel77.  I'll take a look at it this evening.

include/dao/dao.js
```diff
@@ -512,9 +458,8 @@ module.exports = function DAOModule(pb) {
 
                 item.object_type = collection;
                 DAO.updateChangeHistory(item);
-                if (item._id) {
-                    batch.find({_id: item._id})
-                        .updateOne(options.replace ? item : {$set: item });
+                if (item[DAO.getIdField()]) {
```
brianhyder: within the DAO prototype it is safe to refer to the raw ID field ```_id```.

include/service/entities/site_query_service.js
```diff
@@ -259,19 +259,68 @@ module.exports = function SiteQueryServiceModule(pb) {
...
+    });
+  };
+
+  /**
+   * Deletes all site specific content
+   * @param {Array} collections -  array of collection names
+   * @param {String} siteid - unique site id
+   * @param {Function} callback - callback function
+   */
+  SiteQueryService.prototype.deleteSiteSpecificContent = function (collections, siteid, callback) {
```
brianhyder: Where is this function actually called?

include/service/entities/site_query_service.js
```diff
@@ -259,19 +259,68 @@ module.exports = function SiteQueryServiceModule(pb) {
...
+        return callback(BaseObjectService.validationError([BaseObjectService.validationFailure("siteid", "Invalid siteid")]));
+      }
 
-          cb(err, items);
+      var tasks = util.getTasks(collections, function (collections, i) {
+        return function (taskCallback) {
+          self.delete({site: siteid}, collections[i].name, function (err, commandResult) {
+            if (util.isError(err) || !commandResult) {
+              return taskCallback(err);
+            }
```
brianhyder: you will want a different else so you can catch the case where there is no error but the command result is null.  I don't think this is possible TBH but not positive.  I would just remove the ```|| !commandResult``` and call it a day.

include/service/entities/site_query_service.js
```diff
@@ -259,19 +259,68 @@ module.exports = function SiteQueryServiceModule(pb) {
...
+
+            var numberOfDeletedRecords = commandResult.result.n;
+            pb.log.silly(numberOfDeletedRecords + " site specific " + collections[i].name + " records associated with " + siteid + " were deleted");
```
brianhyder: It is more performant to restructure this as:
```
pb.log.silly("%s site specific %s records associated with %s were deleted", numberOfDeletedRecords,  collections[i].name, siteid);
```
The reason is because in the event that the log level is not "silly" then it is cheaper to just put the parameters on the call stack than it is to do the string concatenation, pass it to the function, then find out the log level will not allow for logging the statement.

include/service/entities/site_query_service.js
```diff
@@ -259,19 +259,68 @@ module.exports = function SiteQueryServiceModule(pb) {
...
+            taskCallback(null, {collection: collections[i].name, recordsDeleted: numberOfDeletedRecords});
+          });
+        };
       });
-    };
+      async.parallel(tasks, function(err, results) {
+        if (pb.util.isError(err)) {
+          pb.log.error(err);
```
brianhyder: Why do you log the error here if you are going to pass it back?

include/service/entities/site_query_service.js
kylebechtel77: It is not. I just didnt catch that. I'll remove it.

include/dao/dao.js
kylebechtel77: Will do, thanks for catching that.

brianhyder:
Thanks @kylebechtel77!!!

Issue #803, opened 2015-11-05
Apply Settings


amirsaber:
In site settings page of admin panel, if I change the option of help to improve pencilblue and then click on apply change button, it will spin forever, but if I refresh it I can see the setting took place.

###browser

https://localhost/actions/admin/site_settings 500 (Internal Server Error)

TypeError: Cannot read property 'length' of null
    at ConfigurationSettingsPostController.BaseControllerModule.BaseController.hasRequiredParams (/app/controllers/base_controller.js:352:56)
    at /app/plugins/pencilblue/controllers/actions/admin/site_settings/configuration.js:33:32
    at /app/controllers/base_controller.js:306:13
    at IncomingMessage. (/app/controllers/base_controller.js:335:13)
    at emitNone (events.js:67:13)
    at IncomingMessage.emit (events.js:166:7)
    at endReadableNT (_stream_readable.js:905:12)
    at doNTCallback2 (node.js:450:9)
    at process._tickDomainCallback (node.js:405:17)

###server

2015-11-05T20:47:32.245Z - error: [4] RequestHandler: An error occurred during controller execution. URL=[POST:/actions/admin/site_settings] ROUTE={"path":"actions/admin/site_settings","pattern":"^/actions/admin/site_settings[/]{0,1}$","path_vars":{},"expression":{},"themes":{"pencilblue":{"POST":{"method":"POST","path":"/actions/admin/site_settings","access_level":4,"auth_required":true,"content_type":"text/html"}}}}
TypeError: Cannot read property 'length' of null
    at ConfigurationSettingsPostController.BaseControllerModule.BaseController.hasRequiredParams (/home/ubuntu/workspace/pencilblue/controllers/base_controller.js:352:56)
    at /home/ubuntu/workspace/pencilblue/plugins/pencilblue/controllers/actions/admin/site_settings/configuration.js:33:32
    at /home/ubuntu/workspace/pencilblue/controllers/base_controller.js:306:13
    at IncomingMessage.<anonymous> (/home/ubuntu/workspace/pencilblue/controllers/base_controller.js:335:13)
    at emitNone (events.js:67:13)
    at IncomingMessage.emit (events.js:166:7)
    at endReadableNT (_stream_readable.js:893:12)
    at doNTCallback2 (node.js:429:9)
    at process._tickDomainCallback (node.js:384:17)

amirsaber:
[...]

PR #805, opened 2015-11-06
Issue #803 if result type is boolean and it's been set to false

amirsaber:
I've noticed that in the get prototype of SimpleLayeredService, if the result value is something boolean and false the code will not reach line 69 so the entity will be sent as null while it's not. To catch it I check the type of result and if it's boolean and false it will consider it as an answer and set it as the entity. I think it might be a good idea to start a community forum, a slack or something so we can discuss more of issues and how to handle them and start writing some frontend testing too.

### DIFF
--- a/include/service/simple_layered_service.js
+++ b/include/service/simple_layered_service.js
@@ -65,7 +65,7 @@ module.exports = function SimpleLayeredServiceModule(pb) {
                         return;
                     }
 
-                    if (result) {
+                    if (result || (typeof result === 'boolean' && result === false)) {
                         resultNotFound = false;
                         entity         = result;
                     }
--- a/plugins/pencilblue/controllers/actions/admin/site_settings/configuration.js
+++ b/plugins/pencilblue/controllers/actions/admin/site_settings/configuration.js
@@ -43,7 +43,7 @@ module.exports = function(pb) {
                 if(util.isError(data)) {
                     cb({
                         code: 500,
-                        content: pb.BaseController.apiResponse(pb.BaseController.API_FAILURE, self.ls.get('ERROR_SAVING'), result)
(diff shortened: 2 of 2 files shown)


amirsaber:
Also in configuration.js of site_setting, in line 46 variable result has been written as the third parameter of apiResponse while it's not defind, I believe it meant to be data instead of result.

JahlomP:
I agree with you @amirsaber Slack will be a very good idea.

brianhyder:
@amirsaber, thank you for the fix.  It is much appreciated.

We are a small group so we try to limit the number of places that we are watching for discussions and feature requests.  This is why we rely so heavily on the issues page and attempt to keep the items tags with the milestones that they will be implemented in.

Front end testing would be a nice to have.  It is definitely something that we are interested in but do not have the bandwidth for at the moment.

blakecallens:
I would love for someone from the community to help me with the heavy
lifting of some protractor or karma testing setup.

JahlomP:
Hi @blakecallens. I've been away for sometime now but could you give me some directions so I help out with the tests?

blakecallens:
I'll have to set up a foundation first. Ill reach out when thats done.
Probably Q1 2016.

JahlomP:
K

amirsaber:
@brianhyder that make sense. Thanks for your prompt replies, that's so valuable.

@blakecallens I also can help in the testing process. It will be great if you come up with some foundation of how people can contribute, I've seen you are asking for plug ins and templates and I am working on that. I am creating a RTL template of Portfolio template for RTL languages. I believe providing a structure of how Pencilblue backend works will be also helpful.

TBH, I do like pencilblue a lot. You did a great job. By having the custom objects and some plugins it can even be a full CRM other than being a CMS or even ERP. There is a lot to do and I am excited to help as much as I can.

PR #810, opened 2015-11-07
added functionality to generate article url which fixes #464

### LINKED ISSUE #464: article url generation
I think urls of articles should be generated automatically or by clicking a button specifically for that like checking the availability of a url, to prevent manually adding it since some articles could have  lengthy urls

JahlomP:


plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
```diff
@@ -35,6 +36,21 @@
 			});
 		};
 
+		$scope.generateUrl = function() {
+			$scope.article.url = $scope.article.headline.replace(/\s+/g, '-').toLowerCase();
```
blakecallens: This can be it's own function that returns the generated URL, so we don't need to have the regex defined twice.

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
```diff
@@ -35,6 +36,21 @@
...
+			if($scope.article.url){
+				$scope.urlGenerated = true;
+			} else {
+				$scope.urlGenerated = false;
+			}
+		};
+
+		$scope.$watch('article.url', function() {
```
blakecallens: ```$scope.$watch('article.url', function(newVal, oldVal)...```

Then you can check if the newVal is different than the oldVal before doing a string compare on the generated URL.

plugins/pencilblue/templates/admin/content/articles/article_form.html
```diff
@@ -7,12 +7,26 @@
 		<fieldset>
 			<div class="tab-content">
 				<div class="tab-pane active" id="content">
+					<div class="form-group" ng-class="{'has-error': !isFieldValid(articleForm.headline)}">
```
blakecallens: Why did you move this above the URL?

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
```diff
@@ -35,6 +36,21 @@
 			});
 		};
 
+		$scope.generateUrl = function() {
+			$scope.article.url = $scope.article.headline.replace(/\s+/g, '-').toLowerCase();
+			if($scope.article.url){
+				$scope.urlGenerated = true;
+			} else {
```
blakecallens: We'll never have this hit, if the headline is valid. Also, I would like to make sure the URL is unique/valid. You can make a call to the check function at this point to make sure of that.

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
JahlomP: Will work on that soon

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
JahlomP: Will work on this soon

plugins/pencilblue/templates/admin/content/articles/article_form.html
JahlomP: Since url will be generated from the headline, it is only natural to first define the headline then use it in url generation.

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
JahlomP: Will do the url check but what do you mean by "We'll never have this hit, if the headline is valid."?

plugins/pencilblue/templates/admin/content/articles/article_form.html
blakecallens: I'm fine with that, but it'll need to change on the page form too, for consistency.

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
blakecallens: The generate button is only active if there's a valid headline, which means that a url will always be generated.

plugins/pencilblue/templates/admin/content/articles/article_form.html
JahlomP: K

plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
JahlomP: So should I use the length option?

blakecallens: ```if($scope.article.url && $scope.article.url.length)```

protects against both null and empty string

JahlomP: OK. Thanks.

On Tue, Nov 10, 2015 at 7:01 PM Blake Callens <notifications@github.com>
wrote:

> In
> plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
> <https://github.com/pencilblue/pencilblue/pull/810#discussion_r44448770>:
>
> > @@ -35,6 +36,21 @@
> >  			});
> >  		};
> >
> > +		$scope.generateUrl = function() {
> > +			$scope.article.url = $scope.article.headline.replace(/\s+/g, '-').toLowerCase();
> > +			if($scope.article.url){
> > +				$scope.urlGenerated = true;
> > +			} else {
>
> if($scope.article.url && $scope.article.url.length)
>
> protects against both null and empty string
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pencilblue/pencilblue/pull/810/files#r44448770>.
>
-- 
Best Rregards,
Precious Jahlom Agboado,
www.twitter.com/jahlomp
+233 54 0915 108.

JahlomP:
Done. Please review.

brianhyder:
@JahlomP @blakecallens What is the status on this?  I'd like to merge it into 0.4.2 early this week (Monday or Tuesday).  Thanks.

JahlomP:
Working on it now

JahlomP:
Fixed #842

JahlomP:
@blakecallens review.

blakecallens:
@JahlomP before I can fully review, I'll need a couple things:

1. This pull request needs to be based off of a fork from our repository. That way I can clone your branch that has the changes and test locally. You'll probably have to reopen a new pull request, but it'll make it much easier for me to review.
2. Can you update this pull request to go with the 0.5.0 branch? We're ready to tag 0.4.2 and I think this is a big enough change that it should go in the next major release.

JahlomP:
OK. I agree.

PR #811, opened 2015-11-07
trying to add featured image functionality which is common in most pu…

JahlomP:
…blishing site


brianhyder:
New fields are not automatically accepted by the new ArticleServiceV2.  They must be validated and accounted for in the format, merge, and validate steps.  When the legacy controllers are converted this approach will not work as expected.

brianhyder:
The legacy controllers (new_article.js & edit_article.js) must first be converted over to leverage the new ArticleServiceV2 before this can be merged.

JahlomP:
K

Issue #18, opened 2014-06-30
[Navigation] Editing an Existing Nav Item Re-Orders it to the end of the navigation


blakecallens:
The ordering should be preserved.

JahlomP:
which portion of pb takes care of persisting the order nav items?

brianhyder:
SectionService if I recall correctly

brianhyder:
Big thanks to @amirsaber for the fix.

PR #813, opened 2015-11-09
Fix navigation reordering problem issue #18

amirsaber:
I just had a spare time and I fixed this problem. To do so I added a new function which checkes if the section already exist in the map and if so it return it's index, whether it's a parent node or child node. and then the part that we push into sectionMap, I changed it into checking that if the index exist, splice it and push that into it's previous place.

### DIFF
--- a/include/service/entities/section_service.js
+++ b/include/service/entities/section_service.js
@@ -150,6 +150,40 @@ module.exports = function SectionServiceModule(pb) {
         }
         return orphans;
     };
+    
+        /**
+     *
+     * @private
+     * @method getSectionMapIndex
+     * @param {String} sid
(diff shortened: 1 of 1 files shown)


include/service/entities/section_service.js
```diff
@@ -150,6 +150,40 @@ module.exports = function SectionServiceModule(pb) {
...
         return orphans;
     };
+    
+        /**
+     *
+     * @private
+     * @method getSectionMapIndex
+     * @param {String} sid
+     * @param {Array} sectionMap
+     */
```
brianhyder: Would you please add the ```@return {Object}```?

brianhyder:
The changes look good! Thank you for the contribution.

PR #827, opened 2015-11-11
updates to the admin panel.

kylebechtel77:
This makes the manage sites panel searchable and sortable by date, site name and hostname. This makes it more managable to use when you have several hundred sites.


brianhyder:
@blakecallens When you are good with this PR can you merge?

PR #829, opened 2015-11-11
cleanup and change depricated pb.plugins to new pb.PluginService in jobs/plugins

amirsaber:
I just went though plugin service and while studying it I cleaned it up from js warnings and pb warnings.

### DIFF
--- a/include/service/jobs/plugins/plugin_uninstall_job.js
+++ b/include/service/jobs/plugins/plugin_uninstall_job.js
@@ -44,7 +44,7 @@ module.exports = function PluginUninstallJobModule(pb) {
 
         //initialize
         this.setParallelLimit(1);
-    };
+    }
     util.inherits(PluginUninstallJob, pb.PluginJobRunner);
 
     /**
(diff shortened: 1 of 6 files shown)


include/service/jobs/plugins/plugin_uninstall_job.js
```diff
@@ -99,6 +99,7 @@ module.exports = function PluginUninstallJobModule(pb) {
      */
     PluginUninstallJob.prototype.getWorkerTasks = function(cb) {
         var self = this;
+        var pluginService = new pb.PluginService();
```
brianhyder: Instead of declaring this here.  I would use the instance that is created in prototype constructor.  You should be able to reference an instance of PluginService via ```this.pluginService```

include/service/jobs/plugins/plugin_uninstall_job.js
```diff
@@ -157,7 +158,7 @@ module.exports = function PluginUninstallJobModule(pb) {
             function(callback) {
                 self.log('Attemping to remove theme settings');
 
-                pb.plugins.themeSettingsService.purge(pluginUid, function (err, result) {
+                pluginService.themeSettingsService.purge(pluginUid, function (err, result) {
```
brianhyder: I would update this to use the ```self.pluginService``` instance

brianhyder: The same is true for the other jobs.  They all inherit from ```plugin_job_runner.js```

amirsaber:
There you go @brianhyder :smile:

Issue #836, opened 2015-11-12
[0.5.0] Theme Setting returns 404 on sub-sites


amirsaber:
In 0.5.0, when I go on theme and click on settings of a theme(sample for example) it return 404 page. Is it a known bug?

amirsaber:
It only happens on theme settings on a created site, not on the global. For the same plugin, Theme settings and Plugin settings works fine on global, but Theme settings return 404 on a created site.

blakecallens:
@kylebechtel77 would your team be able to take a look at this?

cchatham:
@blakecallens @brianhyder we are looking into this. @btidwell is looking

blakecallens:
Much appreciated.

amirsaber:
I just noticed that in simple_layered_service.js line 61 return null in case that instance.services[i] is ThemeSettingService and returns correct settings when instance.services[i] is PluginSettingService.

btidwell:
I have found the issue.  In pluginService.js line 616 is using standard pb.DAO() without any site context.  This occurs on plugin install when we call "resetThemeSettings()". So it's saving the settings to the global scope.  As a result the controller for the theme settings page responds with a 404 because theme settings are null.

blakecallens:
Thanks @btidwell!

PR #838, opened 2015-11-12
use site query service instead of DAO to add site context to saving plugin theme settings

btidwell:


amirsaber:
Thanks @btidwell

brianhyder:
Fixes #836

Issue #860, opened 2015-11-16
custom object service deleteTypeById does not accept objectID


amirsaber:
In this function https://github.com/pencilblue/pencilblue/blob/0.5.0/include/service/entities/custom_object_service.js#L952
the definition shows that it accepts id as string and objectid but here:
https://github.com/pencilblue/pencilblue/blob/0.5.0/include/service/entities/custom_object_service.js#L952
it use a validation service which only accepts id strings, so I think this line should be changed to
```
        if (!pb.validation.isId(id, true)) {
```
so it accepts ObjectId and String both.

PR #861, opened 2015-11-16
change so it accepts ObjectId issue #860

amirsaber:


include/service/entities/custom_object_service.js
```diff
@@ -955,7 +955,7 @@ module.exports = function CustomObjectServiceModule(pb) {
             options = {};
         }
 
-        if (!pb.validation.isIdStr(id, true)) {
+        if (!pb.validation.isIdStr(id.toString(), true)) {
```
brianhyder: You should be able to change this to ```if (!pb.validation.isId(id, true)) {```.  Otherwise you risk a null reference error

blakecallens: Then it won't accept strings. I think we need to have both checks here.

brianhyder: That is incorrect.  The ```ValidationService.isId``` function checks both Strings and ObjectID objects.  ```ValidationService.isIdStr``` adds the additional constraint that the value must also be a string as well as parseable into an ObjectID object.  If you look at the implementation of the ```isIdStr``` function you'll see that it validates the value as a string then calls the ```isId``` function.

blakecallens: Ok. Then I was totally wrong in my conversation with @amirsaber today.

brianhyder: No worries, I'm late to the party.  I'll catch up on Gitter tonight.

amirsaber:
Does it look good now?

brianhyder:
All good!

PR #886, opened 2015-12-02
Plugin Init improvements

btidwell:
Add cache for use in plugin initialization, add root node_modules path as a fallback for dependency check, and make use of details.json file for dependency checking instead of DB.

### DIFF
--- a/include/service/entities/plugin_service.js
+++ b/include/service/entities/plugin_service.js
@@ -1046,8 +1046,8 @@ module.exports = function PluginServiceModule(pb) {
 
         async.series(tasks, function(err, results) {
             //cleanup on error
-            if (util.isError(err) && details && details.uid) {
-                delete ACTIVE_PLUGINS[details.uid];
+            if (util.isError(err)) {
+                delete ACTIVE_PLUGINS[plugin.uid];
             }
 
             //callback with final result
@@ -1293,7 +1293,6 @@ module.exports = function PluginServiceModule(pb) {
 
         if(cached_plugin) {
             details = cached_plugin.details;
-            plugin.dependencies = details.dependencies;
             return tasks;
         }
         PLUGIN_INIT_CACHE[plugin.uid] = plugin;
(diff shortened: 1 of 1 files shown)


include/service/entities/plugin_service.js
```diff
@@ -1035,45 +1042,74 @@ module.exports = function PluginServiceModule(pb) {
 
         pb.log.debug("PluginService:[INIT] Beginning initialization of %s (%s)", plugin.name, plugin.uid);
 
+        var tasks = this.getInitTasksForPlugin(plugin);
+
+        async.series(tasks, function(err, results) {
+            //cleanup on error
+            if (util.isError(err) && details && details.uid) {
+                delete ACTIVE_PLUGINS[details.uid];
```
brianhyder: The ```details``` variable is not defined anywhere.  You would also not need the check for the uid in the ```if``` statement because if you make it to here then you have a plugin UID available on the ```plugin``` variable.

include/service/entities/plugin_service.js
```diff
@@ -1202,63 +1241,63 @@ module.exports = function PluginServiceModule(pb) {
...
+        ];
 
-            //callback with final result
-            var success = !util.isError(err);
-            pb.log.info('PluginService: Initialized plugin %s (%s). RESULT=[%s]', plugin.name, plugin.uid, success);
-            cb(err, success);
-        });
+        if(cached_plugin) {
+            details = cached_plugin.details;
+            plugin.dependencies = details.dependencies;
```
brianhyder: Since you are returning on the next line do you need to set the details and plugin.dependencies variable?

include/service/entities/plugin_service.js
```diff
@@ -1277,28 +1316,53 @@ module.exports = function PluginServiceModule(pb) {
...
                 //verify that the module exists and its package definition is available
                 var packagePath = path.join(PluginService.getPluginsDir(), plugin.dirName, 'node_modules', keys[i], 'package.json');
-                var modulePath = packagePath;
-                fs.exists(modulePath, function(exists) {
-                    if (!exists) {
-                        pb.log.warn('PluginService: Plugin %s is missing dependency %s', plugin.name, keys[i]);
-                        
-                        hasDependencies = false;
-                        return callback(null, false);
+                var rootPackagePath = path.join(process.cwd(), 'node_modules', keys[i], 'package.json');
```
brianhyder: Is the assumption that the node_modules directory will always be at the directory where PB is installed?  If so then it might be better to force that assumption: ```path.join(pb.config.docRoot, 'node_modules', keys[i], 'packag.json');```

Either way this implementation is fine as is.

include/service/entities/plugin_service.js
```diff
@@ -1277,28 +1316,53 @@ module.exports = function PluginServiceModule(pb) {
...
+                            pb.log.warn('PluginService: Plugin %s is missing dependency %s', plugin.name, keys[i]);
+                        }
+                        if(!result.satisfied) {
+                            pb.log.warn('PluginService: Plugin %s has incorrect dependency version %s for %s', plugin.name, package.version, keys[i]);
+                        }
+                        if(result.satisfied) {
+                            hasDependencies = true;
+                        }
+                        callback(err, result.satisfied);
+                    });
```
brianhyder: Take a look at ```async.some()```, docs are [here](https://github.com/caolan/async#some).  Should allow you to prevent the nesting.  You can just move the logic that outputs the warning logic to the ```checkModulePath``` function.

brianhyder:
@btidwell Great job on this.  It looks good.  Just had a couple of comments (inline) and one last question:

If a plugin is leveraging a dependency from the PB core can it still call ```PluginService.require``` to load that module?

include/service/entities/plugin_service.js
btidwell: The details yes.  It will throw a null ref during execution of the tasks as details variable is accessed.  The plugin.dependencies no.  I did this before I broke the tasks into two groups.  This is done only for dependency version matching (the version saved to DB might not match the version in details.json).  This is taken care of in the site_independant_tasks.  

I'll remove line 1296.

btidwell:
@brianhyder changes per your comments have been made.

brianhyder:
+1

PR #887, opened 2015-12-03
Locale management

wesanderson303:
Locales can be selected from the site create / edit console


plugins/pencilblue/controllers/admin/sites/site_form.js
```diff
@@ -16,7 +16,6 @@
  */
```
wesanderson303: Hey @brianhyder, this file is 2-space tabbed rather than the conventional 4. I left it 2-space tabbed for now so the PR wasn't spammy. Shall I convert it to 4-space tabs before we're done with this PR?

wesanderson303:
@brianhyder I think this PR's ready to roll. Please give it a once over with feedback when you can. Thanks!

plugins/pencilblue/controllers/admin/sites/site_form.js
brianhyder: I wouldn't worry about it.  Not a big deal.

plugins/pencilblue/controllers/admin/sites/site_form.js
```diff
@@ -75,22 +75,36 @@ module.exports = function SiteFormModule(pb) {
...
   };
 
   function setupAngularObj(self, options, cb){
     var isNew = options.isNew,
       display = options.display,
       host = options.host,
+      supportedLocales = {},
+      savedLocales = options.savedLocales || {},
+      selectedLocales = [],
+      defaultLocale = options.defaultLocale || "",
```
brianhyder: Instead of defaulting to none can we force a default through: ```Localization.getDefaultLocale()```?

plugins/pencilblue/controllers/admin/sites/site_form.js
```diff
@@ -75,22 +75,36 @@ module.exports = function SiteFormModule(pb) {
...
       isActive = options.isActive,
       uid = options.uid;
 
+    for (var locale in pb.Localization.supportedLookup) {
```
brianhyder: Can we use ```Localization.getSupported()``` instead?  It will return an array of locale strings. Then you can simplify things a bit through:

```
selectedLocales = pb.Localization.getSupported().filter(function(locale) {
  return !!savedLocales[locale];
});
```

I'm not sure I understand the difference between savedLocales and selected locales.

brianhyder:
@blakecallens Would you mind taking a quick peek at the UI portion please?

cchatham:
@wesanderson303 I also forgot to tell you to include locales on the global site object when the sites are loaded on startup. I will show you on Monday.

plugins/pencilblue/controllers/admin/sites/site_form.js
wesanderson303: Apologies if this is a bit confusing. I tried to make variables that make sense in the context they're being used. "savedLocales" is populated from the "supportedLocales" object pulled from the DB. "selectedLocales" is used to populate the drop-down and thus needs to be an array. Basically, one's an object, one's an array.

I use this iteration over Localization's supportedLookup to accomplish a few goals:
1. Filter out locales previously saved that may no longer be valid
2. Populate the "supportedLocales" object with true/false values pulled from "savedLocales"(DB)
3. Populate "selectedLocales" with all valid "savedLocales"(DB)

I could use the approach above, but I'd still have to iterate over Localization's supportedLocales to populate the supportedLocales object.

plugins/pencilblue/templates/admin/sites/site_form.html
```diff
@@ -12,12 +12,29 @@
...
                     </div>
+                    <div class="form-group" ng-class="{'has-error': !isFieldValid(siteForm.supportedLocales)}">
+                        <label>Supported Locales</label><br>
+                        <label ng-repeat="(locale, checked) in supportedLocales" style="display:block;">
+                            <input type="checkbox"
+                                   name="supportedLocales"
+                                   value="{{checked}}"
+                                   ng-checked="checked"
+                                   ng-click="updateSupportedLocales(locale)"
+                                    >{{locale}}
```
blakecallens: ```ng-bind="locale"``` is faster and won't show {{locale}} before it renders

brianhyder:
@wesanderson303 There are a couple of merge conflicts.  My apologies.  once those get resolved I'll merge pronto.

wesanderson303:
@brianhyder Sorry about that Brian. Should be good to go now.

PR #889, opened 2015-12-07
BaseObjectService on insert fix

btidwell:
Data being saved through base object service via insert is orphaned due to passing an empty object to onObjectRetrieved function

### DIFF
--- a/include/service/base_object_service.js
+++ b/include/service/base_object_service.js
@@ -537,7 +537,7 @@ module.exports = function(pb) {
         else {
             
             //we are creating a new object so pass it along
-            onObjectRetrieved(null, {});
+            onObjectRetrieved(null, dto);
         }
     };
     


btidwell:
@brianhyder found a bug in BaseObjectService for inserting new data.  When inserting using the save method, only the base fields (dates, id, and site) are saved.

include/service/base_object_service.js
```diff
@@ -537,7 +537,7 @@ module.exports = function(pb) {
         else {
             
             //we are creating a new object so pass it along
-            onObjectRetrieved(null, {});
+            onObjectRetrieved(null, dto);
```
brianhyder: This is not a bug.  It works as designed.  Passing the DTO here will allow for fields to be passed through without explicitly converting them over to the object that will be persisted.

Each service that extends from BaseObjectService must register a "merge" event so that the properties that are desired on the persisted entity are explicitly assigned from the DTO. 

For an example see [here](https://github.com/careerbuilder/pencilblue/blob/master/include/service/entities/content/article_service_v2.js#L602) in the ArticleServiceV2.  Also take a peek at the registration step on line 602.  

I know it seems weird at first but it prevents sites from maliciously adding properties to objects when persisted back through the API.  It also forces developers to be conscious of the data that is being handed back.

btidwell: oh ok.  Thanks for the clarification. :)

PR #892, opened 2015-12-08
Issue/891

brianhyder:


brianhyder:
I'm an idiot and submitted against the wrong branch

PR #895, opened 2015-12-09
Added unix socket support, some rather run NodeJS behind a well confi…

vasilevich:
…gured NGINX/APACHE server, and rather not expose it directly on a TCP PORT, moreover there might be a perfomance gain by avoiding the uneeded TCP IP overhead.

### DIFF
--- a/include/config.js
+++ b/include/config.js
@@ -135,12 +135,13 @@ var BASE_CONFIG = {
 
     //The hostname or IP address that the web server instance will bind to
 	siteIP:   '0.0.0.0',
-
+	
     //The primary port to listen for traffic on.  Some environment such as
     //heroku force you to use whatever port they have available.  In such cases
     //the port is passed as an environment variable.
 	sitePort: process.env.port || process.env.PORT || 8080,
-
+	//The path to an unix socket.
+	siteUnixSocketPath : "/var/run/pencilblue/pb.sock",
     //the absolute file path to the directory where installation lives
 	docRoot:  Configuration.DOCUMENT_ROOT,
 
@@ -629,6 +630,12 @@ var BASE_CONFIG = {
             chain: null
         },
 
+		
+		//allow pencilblue to listen on a TCP IP port  pb.config.siteRoot+pb.config.sitePort
+		tcpip:true,
+		//allow pencilblue to create an UNIX socket
+		unixsocket:true,
+		
         //when non-empty, a header (X-POWERED-BY) will be added to each outgoing
         //response with "PencilBlue".  Cheesy but it helps the BuiltWith tools
         //of the world kep track of who uses what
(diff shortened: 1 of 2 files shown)


vasilevich:
a tiny bug fix, the unix socket file created, had no permission to write/read to, so now I chmod both the folder and the resulting file to 777, you may suggest a different chmod permission.

pencilblue.js
```diff
@@ -221,11 +221,42 @@ function PencilBlue(config){
                 cb(err, false);
             };
             pb.server.once('error', onServerStartError);
-            pb.server.listen(pb.config.sitePort, pb.config.siteIP, function() {
-                pb.log.info('PencilBlue: %s running at site root [%s] on port [%d]', pb.config.siteName, pb.config.siteRoot, pb.config.sitePort);
-                pb.server.removeListener('error', onServerStartError);
-                cb(null, true);
-            });
+			if(pb.config.server.unixsocket==true){
```
blakecallens: Tab spacing is out of line with the rest of the file.

pencilblue.js
```diff
@@ -221,11 +221,42 @@ function PencilBlue(config){
...
+							pb.log.info('PencilBlue: %s running at site at unix socket path:[%s]', pb.config.siteName, pb.config.siteUnixSocketPath, 0);
+							pb.server.removeListener('error', onServerStartError);
+							//give our new socket read/write permission, maybe only read is required.
+							fs.chmodSync(pb.config.siteUnixSocketPath, 0777);
+							cb(null, true);
+						});
+					});
+				}
+				//maybe you want to use both, http and unix same time.
+				if(pb.config.server.tcpip==true){
```
blakecallens: This needs to be the default/fallback, without requiring a tcpip variable, for two reasons.

1. Without it, this would be a breaking change for all current installations.
2. Most users will want to use tcpip. The unixsocket settings are the use case exception.

include/config.js
```diff
@@ -629,6 +630,12 @@ var BASE_CONFIG = {
             chain: null
         },
 
+		
+		//allow pencilblue to listen on a TCP IP port  pb.config.siteRoot+pb.config.sitePort
+		tcpip:true,
+		//allow pencilblue to create an UNIX socket
+		unixsocket:true,
```
blakecallens: We can't assume that this is only being run on unix. It needs to be something that's turned on.

blakecallens:
Thanks for the submission @vasilevich. With 0.5.0RC coming this month, we have a code freeze on 0.4.x. Would you mind resubmitting this against the 0.5.0 branch and testing to make sure it works with the new multi-site feature?

vasilevich:
sure, either way, I have it fully tested and working in my arch linux system.
I use it with NGINX configured as a front end server serving static content via SSL
while pencilblue is configured as an app server responding to dynamic requests alone via the UNIX SOCKET.
I thought you might want to add NGINX/APACHE configuration examples in your git source such as the nginx config file I attached
[duckgamer.txt](https://github.com/pencilblue/pencilblue/files/57867/duckgamer.txt)
this is an example of using ssl and the socket path.

include/config.js
vasilevich: yes exactly, thats why I made it a variable, so you can set it to true/false, the true is just a demonstration that both can be used at the same time.

PR #896, opened 2015-12-10
0.5.0

vasilevich:
Added UNIX SOCKET support for users that wish to put their pencilblue installation behind another webserver such as NGINX or APACHE.

### DIFF
--- a/include/config.js
+++ b/include/config.js
@@ -110,7 +110,7 @@ Configuration.getBaseConfig = function(multisite) {
         //The root of the site.  This host part should ALWAYS match the value of
         //the siteIP
         siteRoot: 'http://localhost:8080',
-
+		
         //The hostname or IP address that the web server instance will bind to
         siteIP:   '0.0.0.0',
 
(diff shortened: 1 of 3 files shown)


brianhyder:
@vasilevich We will be declining this PR. 

1. The code does not actually run.  There is a syntax error.  Please test all code that is submitted and run the unit test suite.  I understand the failure is probably due to the merge when you submitted the PR for the 0.5.0 branch.
2. We haven't had any other requests for this functionality.  In this case we want to minimize the amount of code that we are responsible for maintaining. 

This afternoon I spent some time looking at how the HTTP server was created.  It was seriously lacking in flexibility.  I refactored it to support the ability to create your own server setups.  Below is a code sample that you could use to get up and going in 0.5.0.

CustomServerInitializer - can be part of a plugin
```
//dependencies
var fs = require('fs');
var ServerInitializer = require('include/http/server_initializer.js');//note that this path will have to change

function CustomServerInitializer(pb){
  this.config = pb.config;
}
util.inherits(CustomServerInitializer, ServerInitializer);

CustomServerInitializer.prototype.startServer = function(server, port, ip, cb) {
  //put your code to create the socket - you can access the config because you set it in the constructor
  //put your call to server.listen here
  //don't forget to call back.
};

module.exports = CustomServerInitializer
```

Use the the config.js file to set the initializer or your custom properties
```
{
  server: {
    initializer: require('path/to/custom_server_initializer.js')
  }
}
```

PR #898, opened 2015-12-10
Fixes a wysiwyg media preview bug in MT mode because site data was no…

andparker:
…t being passed to the MediaService constructor.


andparker:
In MT mode, this was causing a infinite loop of failed(400) api calls to the get_media_embed service to render a media preview in the content WYSIWYG.

PR #904, opened 2015-12-14
Edit 301 redirects

wesanderson303:
@brianhyder Wanted to get a conversation going about the best way to redirect to sites with edited domains. Please don't merge this PR. I'll re-commit once we get our 0.5.0 fork properly merged.


include/http/request_handler.js
```diff
@@ -144,6 +145,87 @@ module.exports = function RequestHandlerModule(pb) {
...
+                        uid: uid};
+        var redirectHost = pb.DocumentCreator.create('redirect_hosts', redirect);
+        dao.save(redirectHost, function(err, result) {
+            if (pb.util.isError(err)) {
+                pb.log.error("ERROR: Failed to save off redirectHost ERR[" + err.stack + "]");
+
+            }
+        });
+    };
+
```
wesanderson303: TODO: Delete all required redirects when a site is deleted.

include/http/request_handler.js
```diff
@@ -144,6 +145,87 @@ module.exports = function RequestHandlerModule(pb) {
...
+    /**
+     * @static
+     * @method checkForRedirect
+     * @param {String} hostname
+     * @param {Function} cb
+     */
```
wesanderson303: I'd be open to suggestions on the best way to perform the actual redirect

include/service/entities/site_service.js
```diff
@@ -123,6 +142,16 @@ module.exports = function SiteServiceModule(pb) {
...
 
     /**
+     * Get all of the redirect_hosts in the database
+     * @method getRedirectHosts
+     * @param {Function} cb - the callback function
+     */
+    SiteService.prototype.getRedirectHosts = function(cb) {
+      var dao = new pb.DAO();
+      dao.q(REDIRECT_HOSTS, { select: pb.DAO.SELECT_ALL, where: {} }, cb);
+    };
```
wesanderson303: New DB collection: "redirect_hosts". Is this the best way to retrieve this collection?

include/service/jobs/sites/site_create_edit_job.js
```diff
@@ -98,7 +93,13 @@ module.exports = function SiteCreateEditJobModule(pb) {
                         site = pb.DocumentCreator.create('site', mySite);
                     }
 
-                    site.hostname = mySite.hostname || site.hostname;
+                    var hostname = mySite.hostname || site.hostname;
+                    // If this site's hostname has been changed, save off a redirectHost for 301 redirects
+                    if ((site && mySite) && (site.hostname !== hostname) && mySite.uid) {
+                        pb.RequestHandler.saveRedirectHost(mySite.hostname, site.hostname, mySite.uid);
```
wesanderson303: I'm piggy-backing the site save job to update the RequestHandler's redirectHosts object. Is this worth spinning off a separate job?

include/service/entities/site_service.js
brianhyder: I think you would want to store all of the previous hostnames in an array or hash inside of the site object itself.  That way the site itself is aware of its evolution over time.

include/http/request_handler.js
```diff
@@ -144,6 +145,87 @@ module.exports = function RequestHandlerModule(pb) {
...
+            siteObj = RequestHandler.sites[hostname];
+
+        if (siteObj) {
+            return cb(err, siteObj);
+        }
+
+        else {
+            var redirectSite = RequestHandler.redirectHosts[hostname];
+            if (redirectSite) {
+                new pb.SiteService().getAllSites(function (err, results) {
```
brianhyder: You definitely do not want to do this.  A DB call on every request is extremely heavy.  If you have the cached site keep track of its previous hostnames you eliminate the DB call and the possibility of redirecting N number of times until you get to the latest name.  

When a site is "loaded" it should take its array of previous host names and register then in a lookup hash of old hostname to new hostname.  When the request comes in you take the hostname and compare it to the hash of "old" hostnames.  If you get a match then replace the hostname in the request URL and redirect.

This limits the data changes to just the site service's merge function:
```
var dto = ...;
var entity = ...;
entity.prevHostNames = util.isArray(entity.prevHostNames) ? entity.prevHostnames : [];

//when a new hostname is provided (excluding on creation) we add the current hostname to the prev list before assigning the new hostname
if (dto.hostname !== entity.hostname && !!entity.hostname) {
  if (entity.prevHostnames,indexOf(dto.hostname) == -1) {
    entity.prevHostNames.push(entity.hostname);
  }
}
entity.hostname = dto.hostname;
```

include/service/jobs/sites/site_create_edit_job.js
brianhyder: No, but if the previous hostnames are part of the site object then this step is not needed and can be taken care of as part of the call to ```RequestHandler.loadSite(...)```

include/service/entities/site_service.js
wesanderson303: This would require iterating over every active site's old host names to find which active site to redirect to. That's a tad more expensive than grabbing an active site uid from a hash array of hostnames that need to be redirected.

include/http/request_handler.js
wesanderson303: Ah, I like this better. Sorry, I didn't know getAllsites() hit the DB each time. So in summary - 
On site retrieval, generate and populate a memory-cached object of hostnames to redirect to onSessionRetrieved()? No need for a new DB collection

include/service/entities/site_service.js
brianhyder: You and I are on the same page but saying different things.

Let's say you have a site object:
```
{
  hostname: "test1.localhost:8080",
  prevHostNames: [
    "old.localhost:8080"
  ]
}
```

It gets passed to ```RequestHandler.loadSite(site);```

```
function(site) {
  RequestHandler.sites[site.hostname] = { active: site.active, uid: site.uid, displayName:  site.displayName, hostname: site.hostname };
  site.prevHostNames.forEach(function(oldHostname) {
    RequestHandler.siteRedirects[oldHostname] = site.hostname;
  }
}
```

You lookup for each request is constant time and you avoid the iteration.

include/http/request_handler.js
brianhyder: Sounds about right.

wesanderson303: I like it. I'll play around with this approach tomorrow. Thanks for the feedback!

wesanderson303:
@brianhyder Would you mind reviewing this PR? I think my changes were in line with your recommendations.

brianhyder:
Yes, I can do that. I'll add it to my list of TODOs.

include/service/jobs/sites/site_create_edit_job.js
```diff
@@ -96,7 +93,24 @@ module.exports = function SiteCreateEditJobModule(pb) {
...
+                        for (var i=0; i< site.prevHostnames.length; i++) {
+                            if (site.prevHostnames[i] == newHostname) {
+                                site.prevHostnames.splice(i, 1);
+                            }
+                        }
+                        site.prevHostnames.push(prevHostname);
+                        pb.RequestHandler.redirectHosts[prevHostname] = newHostname;
+                        pb.RequestHandler.sites[prevHostname] = null;
+                    }
+                    site.hostname = newHostname || prevHostname;
```
brianhyder: This logic would be better suited to live in the ```SiteService.merge``` function.  It would isolate the assignment of fields to the service that owns the object.

Depending on how out of scope you'd like to get... You don't actually need to call to load the site via the call to ```dao.loadById```.  The save method will do this automatically if the ```_id``` property is set on "mySite".  All you have to do is pass the raw ```mySite``` object to save and let the SiteService handle the rest.

include/service/jobs/sites/site_create_edit_job.js
```diff
@@ -96,7 +93,24 @@ module.exports = function SiteCreateEditJobModule(pb) {
                         site = pb.DocumentCreator.create('site', mySite);
                     }
 
-                    site.hostname = mySite.hostname || site.hostname;
+                    var prevHostname = site.hostname;
+                    var newHostname = mySite.hostname;
+                    if (!site.prevHostnames) {
+                        site.prevHostnames = [];
```
brianhyder: Move this if check and assignment to the ```SiteService.format``` method.  On save a number of methods and events get called:

1. The format event is fired.  This event allows for raw DTOs to be properly handled.  i.e. ISO-8601 strings to Date objects, and sanitizing string fields.
2. The object is retrieved from the DB (only when ID is set)
3. The merge event is fired.  This is where the properties from the DTO are merged into the existing entity object.
4. The validate event is fired.  This allows for the whole object to be validated before being persisted.  When errors occur they are handed back in the callback as an error with a ```validationErrors``` property.
5. The before save event is fired. This is there just in case there is a last minute action to perform before persistence.
6. The object is persisted to the DB
7. The after save event is fired.  This is here in case you want to do caching or something along those lines.

include/service/jobs/sites/site_create_edit_job.js
```diff
@@ -96,7 +93,24 @@ module.exports = function SiteCreateEditJobModule(pb) {
...
+                    }
+                    // If this site's hostname has been changed, save off a redirectHost
+                    if ((prevHostname && newHostname) && (prevHostname !== newHostname)) {
+                        // Check for circular hostname references
+                        for (var i=0; i< site.prevHostnames.length; i++) {
+                            if (site.prevHostnames[i] == newHostname) {
+                                site.prevHostnames.splice(i, 1);
+                            }
+                        }
```
brianhyder: Good catch.  I wouldn't have thought about that.  You can simplify a bit:

```
site.prevHostnames = site.prevHostNames.filter(function(hostname) {
  return site.prevHostnames[i] !== newHostname;
})
```

wesanderson303:
@brianhyder Adjusted per your recommendations. Thanks for the feedback. Should be good to go again. Hope you enjoyed the holiday!

include/service/entities/site_service.js
```diff
@@ -289,6 +289,9 @@ module.exports = function SiteServiceModule(pb) {
      */
     SiteService.prototype.editSite = function(options, cb) {
         cb = cb || util.cb;
+        if (!options.prevHostnames) {
+            options.prevHostnames = [];
+        }
```
wesanderson303: @brianhyder Er - don't merge yet please. This check and assignment does nothing. My tests were only working on sites with a prevHostnames array pre-defined. Your previous comment suggested adding this assignment to the SiteService.format method but this does not exist, nor should it? I don't want to register for the BaseObjectService's format event in the SiteService which inherits from the BOS right? I don't want to handle events emitted by myself? Would you mind clarifying a little?

brianhyder: Each service that inherits from BOS should register for and handle some events.  The majority of the time this should be "merge", "format", and "validate".  

You are correct, ensuring that the prevHostnames property is an array should not be performed in the format method.  Instead, it should be in the merge since it is a protected property that is essentially, read-only.  I would do the check in the merge handler.  My apologies.

PR #909, opened 2015-12-15
Redirect 301

wesanderson303:
@brianhyder Don't merge plz. Just a conversation starter until we finish our local 0.5.0 merge. This will redirect with status code 301. Defaults to 302 if no statusCode's provided.

### DIFF
--- a/include/http/request_handler.js
+++ b/include/http/request_handler.js
@@ -100,6 +100,15 @@ module.exports = function RequestHandlerModule(pb) {
     RequestHandler.staticRoutes = {};
 
     /**
+     * Redirect http codes
+     * @static
+     * @property REDIRECT_CODES
+     * @type {Object}
+     */
+    RequestHandler.REDIRECT_CODES = {MOVED_PERMANENTLY: 301,
+                                     FOUND: 302};
+
+    /**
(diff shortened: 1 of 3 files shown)


include/http/request_handler.js
```diff
@@ -91,6 +91,15 @@ module.exports = function RequestHandlerModule(pb) {
     RequestHandler.staticRoutes = {};
 
     /**
+     * Redirect http codes
+     * @static
+     * @property REDIRECT_CODES
+     * @type {Object}
+     */
+    RequestHandler.REDIRECT_CODES = {MOVED_PERMANENTLY: 301,
```
wesanderson303: Is there a better place to place a static for status codes?

brianhyder: Not something we want to keep up with.  I would use: https://www.npmjs.com/package/http-status-codes

brianhyder: We could also add it to the requirements.js file so that all you have to do is: ```pb.HttpStatus.OK```

wesanderson303:
@brianhyder I think this PR's all ready to go. Mind taking a look when you get some time?

PR #912, opened 2015-12-21
fixes #907 - media sizing

### LINKED ISSUE #907: Media image maxHeight not being set due to case mismatch for style attribute
I have checked both 0.4.1 and 0.5.0 and this bug is present.  In wsywig.js, line 154, the maxheight attribute is set with the name "maxheight" (i.e., no uppercase H for height).  In media_service.js, line 520, you check for "maxHeight" style (Capital H for Height).  In stepping through the code, the maxHeight doesn't get set because the passed style attribute doesn't match the case.  Fix should be either in wsywig.js or in media_service.js to match case.  The line numbers are for 0.5.0 version.

blakecallens:


blakecallens:
@thcase I was able to create a pull request for you. Your Visual Studio config is in the branch, though, so you'll have to remove it and recommit before we can merge. Thanks for the help!

brianhyder:
Before this is merged please update the PR to remove and ignore the ```.vscode/launch.json``` file as it is IDE specific.

thcase:
I have updated my repository to remove the .vscode directory and .vscode/launch.json file.  Do I need to issue a pull request?

brianhyder:
@thcase Nope, this is perfect.  Thank you for the fix!

PR #918, opened 2016-01-04
Load site defaults

wesanderson303:


include/http/request_handler.js
```diff
@@ -164,9 +164,9 @@ module.exports = function RequestHandlerModule(pb) {
           uid: site.uid,
           displayName: site.displayName,
           hostname: site.hostname,
-          defaultLocale: site.defaultLocale,
-          supportedLocales: site.supportedLocales,
-          prevHostnames: site.prevHostnames
+          defaultLocale: site.defaultLocale || pb.Localization.getDefaultLocale(),
+          supportedLocales: site.supportedLocales || [],
```
brianhyder: @wesanderson303 Are you sure this works as expected?  I *think* this will always default to English because there are no other allowed locales for the site

brianhyder: It might be cleaner to ensure defaults in the site service and do error checking here.  thoughts?

wesanderson303: Is there something else the defaultLocale should resolve to?


- Wes Anderson


-------- Original message --------

brianhyder:
Yeah, look at how the site service picks the available locales for the global site.

wesanderson303:
It looks like a defaultLocale is already set and pulled from pb.Localization in SiteService.getByUid() for the global site. But this won't help for non-global sites loaded in the request_handler without the or. I'm a little confused by where else we could fallback assign the supportedLocales if not in .loadSite(). Would you mind offering a little more clarity?

brianhyder:
I was thinking about SiteService line ~385 where it loops over loads each site.  You could then ensure that the values are there or set the defaults.  I alluded to how the global site was done.  I meant how the default locales are pulled and set.  This could be the default if the site doesn't not have supported locales or a default locale available.

As long as you all are comfortable with the behavior in the scenario where you have to default it then I'm OK with it.  Let me know how y'all want to proceed.

wesanderson303:
I have no problem moving the default assignments to SiteService. I'll push that momentarily but this concept raises 2 theories I'd like to bounce by you:

1. In reference to the edit_site question earlier, would it be worth running each site through a "scrubber" method, providing a chance to A) Catch null values and assign defaults and B) Giving devs a one-stop-shop for forward-facing site cleanup. In other words, clean up your sites on load instead of edit.

```javascript
//only load the sites when we are in multi-site mode
if (pb.config.multisite.enabled) {
    util.forEach(results, function (site) {
        site = scrubSitePreLoad(site);
        pb.RequestHandler.loadSite(site);
    });
}

SiteService.scrubSitePreLoad = function(site) {
    //Iterate over site and null assign and optional cleanup?
}
```
2. Does the Global site need 'defaultLocale' and 'supportedLocales'? I mean, I know it doesn't need 'prevHostnames', but from a code security standpoint, would it be worth them all looking uniform?

brianhyder:
For number 1 I struggle see the need to include code that will only ever execute for ta small percentage of users.  This code, in theory, has to be tested and maintained.  Since multi-site hasn't officially released I'm weary of introducing code for backward compatibility when it can be fixed by running an update statement against the DB or simply viewing and re-saving the site in the UI.  

On number 2 I think it would be good to add an empty array to the global site for uniformity's sake.

brianhyder:
Looks good!

PR #919, opened 2016-01-05
Fixing error logging

kylebechtel77:
Fixes a missing %s at the end of the file for logging uncaught exceptions.


brianhyder:
Good catch!

PR #920, opened 2016-01-06
Site db tweaks

wesanderson303:
See comments below for justifications and questions


plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
```diff
@@ -53,8 +53,8 @@ module.exports = function DeleteSiteActionModule(pb) {
         var self = this;
         var siteid = self.pathVars.siteid;
         this.service.deleteSingle({where: {uid: siteid}}, function (err, site) {
-            if (util.isError(err)) {
-                return cb(err);
+            if (util.isError(err)|| !site) {
+                err = new Error("Failed to retrieve site from database for deletion");
```
wesanderson303: Per your email, you recommended a 404? I was working with this locally and prefer this flow. Displays a pop-up on site delete. Your call though. I have no problem doing a 404 instead.

plugins/pencilblue/controllers/actions/admin/sites/new_site.js
```diff
@@ -48,7 +48,7 @@ module.exports = function NewSiteActionModule(pb) {
             var selectedLocale = self.body.selectedLocales[i];
             self.body.supportedLocales[selectedLocale] = true;
         }
-        self.body.selectedLocales = undefined;
+        delete self.body.selectedLocales;
```
wesanderson303: An oversite in original push for the locale submission. The original would include a null 'selectedLocales' entry to the db for the site object. This should clean that up for sites going forward, which raises another question:

For existing 'corrupted' sites with null 'site' and 'selectedLocale' entries, is it worth scrubbing those clean inside edit_site?

plugins/pencilblue/controllers/actions/admin/sites/delete_site.js
brianhyder: We would prefer a 404 when the requested site to delete does not exist.  This aligns better with the direction we are taking the newer APIs.

plugins/pencilblue/controllers/actions/admin/sites/new_site.js
brianhyder: My suggestion is no.  0.5.0 has not been officially released so the majority of users wouldn't benefit from it.  You guys will be better off running a one time update against your environments' DB to do the cleanup.  Think of it as a schema change.

brianhyder:
Well done!

PR #923, opened 2016-01-11
0.5.0

wesanderson303:
@brianhyder A few changes to site initiailization: supportedLocales should be an object instead of an array. But doing some digging, I have some concerns. May be un-founded, but wanted your opinion: SiteMapService sets its supportedLocales from localization.js, and I'm not certain it returns an object (see line 666). If it does indeed return an array, it would need some tweaks to become an object instead as that's how it's used elsewhere.


include/localization.js
```diff
@@ -472,7 +472,6 @@ module.exports = function LocalizationModule(pb) {
      * @param {Function} cb
      */
     Localization.init = function(cb) {
-        var supportedLocales = [];
```
wesanderson303: Some of the changes in this PR confuse me. This one for instance has already been merged? (https://github.com/careerbuilder/pencilblue/pull/154)

brianhyder: Yeah, I could have sworn I had already fixed the chunk above.

include/service/entities/site_service.js
```diff
@@ -598,13 +602,16 @@ module.exports = function SiteServiceModule(pb) {
...
+        defaultSupportedLocales[defaultLocale] = true;
         return {
             displayName: pb.config.siteName,
             uid: pb.SiteService.GLOBAL_SITE,
             hostname: pb.config.multisite.enabled ? url.parse(pb.config.multisite.globalRoot).host : url.parse(pb.config.siteRoot).host,
             active: pb.config.multisite.enabled ? false : true,
-            defaultLocale: pb.Localization.getDefaultLocale(),
-            supportedLocales: util.arrayToObj(pb.Localization.getSupported(), function(a, i) { return a[i]; }, function(a, i) { return true; }),
+            defaultLocale: defaultLocale,
+            supportedLocales: defaultSupportedLocales,
```
brianhyder: Agreed. This should be an object.  Using a single locale for the global site isn't what we want to do.  The line that was removed asks localization for an array of Locale objects.  It then runs that through the ```util.arrayToObj``` which allows for converting an array to a hash.  You can run it through the debugger but I'm pretty sure that it provides an object as expected.

wesanderson303: Rockin. That addresses my concern then. So then we're an object everywhere. I'll push that back momentarily

include/localization.js
wesanderson303: Is this just git being funky?

brianhyder: That is my guess or commits got lost on one end but are being pushed back by your fork.  Not sure but seems to be all there between your push and the branch we have.

PR #944, opened 2016-02-15
Autosync plugin settings

wesanderson303:
Creates a config setting: "autoSync" to compare plugin settings' details.json settings against the DB-loaded values. If a discrepancy is detected, the plugin is saved off.

### DIFF
--- a/include/config.js
+++ b/include/config.js
@@ -219,6 +219,7 @@ Configuration.getBaseConfig = function(multisite) {
         settings: {
             use_memory: true,
             use_cache: false,
+            autoSync: false,
 
             //The timeout specifies how long in milliseconds a setting will exist
             //in memory before being flushed.  A value of 0 indicates that the
--- a/include/service/entities/plugin_service.js
+++ b/include/service/entities/plugin_service.js
@@ -560,7 +560,7 @@ module.exports = function PluginServiceModule(pb) {
                 hashedSettings[settingName] = true;
                 if (settings[settingName] === undefined) {
                     discrepancy = true;
-                    formattedSettings.push({name: details.settings[i].name, value: details.settings[i].value});
+                    formattedSettings.push({name: settingName, value: details.settings[i].value});
                 }
                 else {
                     formattedSettings.push({name: settingName, value: settings[settingName]});
--- a/include/service/entities/plugin_service.js
+++ b/include/service/entities/plugin_service.js
@@ -1,19 +1,19 @@
 /*
-    Copyright (C) 2015  PencilBlue, LLC
+ Copyright (C) 2015  PencilBlue, LLC
 
-    This program is free software: you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation, either version 3 of the License, or
-    (at your option) any later version.
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
 
-    This program is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU General Public License for more details.
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
 
-    You should have received a copy of the GNU General Public License
-    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-*/
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
 
 //dependencies
 var fs      = require('fs');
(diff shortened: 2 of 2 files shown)


wesanderson303:
Hey @brianhyder, per our email discussions, I attempted to implement something that could handle plugin settings per site at scale (what we'll need), while attending to the concerns you alluded to. I've tested this here across 59 sites (only some of which needed updating when a discrepancy was detected). Spin up time before implementing this was ~4 seconds. Afterwards, I noticed an increase of ~.5 second.

I'm open to methods of optimizing this, but this felt like the most sound solution for helping us keep our plugin settings in sync with the details.json at scale.

include/service/entities/plugin_service.js
```diff
@@ -1079,7 +1139,13 @@ module.exports = function PluginServiceModule(pb) {
                 if(!cached_plugin || !cached_plugin.details) {
                     return PluginService.loadDetailsFile(PluginService.getDetailsPath(plugin.dirName), function (err, loadedDetails) {
                         details = loadedDetails;
-                        callback(err, !!details);
+                        if (autoSync && !pb.util.isError(err) && details) {
+                            var pluginService = new PluginService({site: site});
```
wesanderson303: A new pluginService with the plugin's site here seemed un-optimized. I need this so that getSettingsKV() works against the PluginService's site in syncSettings. Is there a better way to handle this? One alternative I considered was changing the current PluginService's site before calling syncSettings and then restoring it after the getSettingsKV() call, but that scares me.

include/service/entities/plugin_service.js
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
+
+            // Detect deprecated settings if we haven't already detected a discrepancy
+            if (!discrepancy) {
+                for (var setting in settings) {
+                    if (!hashedSettings[setting]) {
+                        discrepancy = true;
+                        break;
+                    }
+                }
+            }
```
wesanderson303: We brainstormed methods of detecting discrepancies for a good long while and these ~2 loops seemed the most optimized but I'm open to alternatives if you can think of one.

include/service/entities/plugin_service.js
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
      * persisted.
      *
+     * @method syncSettings
+     * @param details The details object to extract the settings from
+     * @param cb A callback that provides two parameters: cb(error, TRUE/FALSE).
+     * TRUE if the settings were successfully cleared and reloaded. FALSE if not.
+     */
+    PluginService.prototype.syncSettings = function(plugin, details, cb) {
+        var self = this;
+        this.getSettingsKV(plugin.dirName, function(err, settings) {
```
wesanderson303: Again, as mentioned below, the PluginService used here's site needs to be mapped to the plugin in question for this to work properly. Ideas to the best way to handle this other than instantiating a new PluginService each time?

brianhyder:
@wesanderson303 I got a chance to go through it.  Design looks solid so well done.  There are some items that we'd like to get cleaned up but nothing major.  I'll start commenting on those.

If you would though, please close the PR and re-open that against the "0.5.1" branch.  We'd like to start putting all new functionality there for the next release.
@cchatham I meant to send you a note about the next release branch and forgot.  My bad.

include/config.js
```diff
@@ -219,6 +219,7 @@ Configuration.getBaseConfig = function(multisite) {
         settings: {
             use_memory: true,
             use_cache: false,
+            autoSync: false,
```
brianhyder: Since this only applies to plugin settings I'd like to eliminate any ambiguity and make it: ```pb.config.plugins.syncSettingsAtStartup```.  I know it is verbose but hopefully there won't be any mistaking its purpose.

brianhyder: If you could, please include a line or two about how it works etc.  This just helps us keep the source documented and ensures that it evolves with the code.

include/service/entities/plugin_service.js
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
+            if (pb.util.isError(err) || !settings) {
+                pb.log.error("PluginService.syncSettings failed to load settings from plugin [" + plugin.dirName + "]");
```
brianhyder:
```
pb.log.error("PluginService: Failed to load settings from plugin [%s]", plugin.uid);
```

include/service/entities/plugin_service.js
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
+                return cb(null, true);
```
brianhyder: You'll want to callback with the err and false.

```return cb(err, false);```

File: include/service/entities/plugin_service.js [hunk 7]
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
+            for (var i = 0; i < details.settings.length; i++) {
+                var settingName = details.settings[i].name;
+                hashedSettings[settingName] = true;
+                if (settings[settingName] === undefined) {
+                    discrepancy = true;
+                    formattedSettings.push({name: settingName, value: details.settings[i].value});
+                }
+                else {
+                    formattedSettings.push({name: settingName, value: settings[settingName]});
+                }
```
brianhyder:
```
details.settings.forEach(function(setting) {
  var settingName = setting.name;
  hashedSettings[settingName] = true;
  var val = settings[settingName];
  if (typeof val === 'undefined') {
    discrepancy = true;
    val = setting.value;
  }
  formattedSettings.push({name: settingName, value: val});
});
```

include/service/entities/plugin_service.js
brianhyder: Double check me but I *think* that at this point all you have to do is verify that the number of items in the ```formattedSettings``` array matches the number in of keys in ```settings```:

My guess if you can eliminate the loop and if statement by:
```
discrepancy = discrepancy || (details.settings.length !== Object.keys(settings));
```

include/service/entities/plugin_service.js
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
+                self.resetSettings({uid: plugin.uid, settings: formattedSettings}, function(err, result) {
+                    if (pb.util.isError(err)) {
+                        pb.log.error("PluginService.syncSettings failed to save off updated settings for " + plugin.dirName);
+                    }
+                    cb(null, true);
+                });
+            }
+            else {
+                cb(null, true);
+            }
```
brianhyder: By inverting the if/else you can avoid having to nest as deeply:

```
//all good
if (!discrepancy) {
  return cb(null, true);
}

self.resetSettings({uid: plugin.uid, settings: formattedSettings}, function(err, result) {
  if (util.isError(err)) {
    pb.log.error("PluginService.syncSettings failed to save off updated settings for " + plugin.dirName);
  }
  cb(err, !util.isError(err));
});
```

include/service/entities/plugin_service.js
```diff
@@ -538,6 +538,65 @@ module.exports = function PluginServiceModule(pb) {
...
+        });
+    };
+
+    /**
+     * Loads the settings from a details object and persists them in the DB.  Any
+     * existing settings for the plugin are deleted before the new settings are
+     * persisted.
```
brianhyder: +1 for the documentation assist!

include/service/entities/plugin_service.js
```diff
@@ -1071,6 +1130,7 @@ module.exports = function PluginServiceModule(pb) {
         var details = null;
         var site = plugin.site || GLOBAL_SITE;
         var cached_plugin = PLUGIN_INIT_CACHE[plugin.uid] || null;
+        var autoSync = pb.config.settings.autoSync || false;
```
brianhyder: It is reasonable to assume that there will be a value for the setting.  I wouldn't default it.

include/service/entities/plugin_service.js
brianhyder: The reality of the situation is that this function is WAY to big and should be moved to its own prototype at some point in the near future.  Until then i can live with it.  However, you are correct.  If you could just add an inline comment with a TODO to address this when that piece is broken out I would appreciate it.

include/service/entities/plugin_service.js
wesanderson303: oo nice catch. I like that. I'll try it out in the morning.

wesanderson303:
@brianhyder Thanks for all the feedback Brian. I'll have it all addressed and re-submitted to 0.5.1 in the morning.

PR #948, opened 2016-02-18
This adds the ability to edit localizations from the admin panel

kylebechtel77:
This PR adds a side nav item to the admin panel that allows you to edit and override localizations via the current file based system. It also makes localizations site and active theme aware. 

This is the order of localization selection: 
1. Site/active theme
2. Site/any theme
3. Active theme
4. Any theme
5. Generics

It also adds a few configuration options:
localization.db = true will set the localization file to use a database instead of files
localization.nav = true will let you put localizations directives in your navigation items
localization.pages = true will let you put localization directives in content pages. 
templateService.useReprocess = true has to be on for .nav and .pages to work properly.


controllers/base_controller.js
```diff
@@ -117,6 +117,9 @@ module.exports = function BaseControllerModule(pb) {
         this.body                = props.body;
         this.localizationService = props.localization_service;
         this.ls                  = this.localizationService;
+        this.ls.activeTheme      = props.activeTheme;
+        this.ls.siteName         = props.siteName;
+        this.ls.site             = props.site;
```
brianhyder: Seems redundant to set these here.  Why not pass them in as options when the localization instance is created in the RequestHandler?

controllers/base_controller.js
```diff
@@ -151,8 +154,26 @@ module.exports = function BaseControllerModule(pb) {
             onlyThisSite: true,
             siteObj: this.siteObj
         };
+        if(pb.config.localization && pb.config.localization.db){
```
brianhyder: Making a call to the DB for every request to load localizations doesn't seem optimal.  Did you consider having a job to sync up the site specific options across the cluster in memory?

controllers/base_controller.js
```diff
@@ -151,8 +154,26 @@ module.exports = function BaseControllerModule(pb) {
...
+            var opts = {
+                where: {_id: self.site}
+            };
+            var queryService = new pb.SiteQueryService({site: self.site, onlyThisSite: true});
+
+            queryService.q("localizations", opts, function (err, result) {
```
brianhyder: it appears that all localization overrides for all sites are stored in a single object.  This isn't optimal because we have to load the document that contains items from all sites which we don't need.

controllers/base_controller.js
```diff
@@ -151,8 +154,26 @@ module.exports = function BaseControllerModule(pb) {
...
+                if (util.isError(err)) {
+                    pb.log.error(err);
+                }
+                if(result && result[0] && result[0].storage)
+                    pb.Localization.storage =  util.deepMerge(result[0].storage,pb.Localization.storage);
```
brianhyder: This step is very redundant.  With every request you are updating keys in storage that were probably just updated by the previous request.  Did you test this to ensure that the merge successfully removed keys?  I'm not sure the merge will remove the deleted overrides if they had previously been merged before.

include/config.js
```diff
@@ -131,6 +131,18 @@ Configuration.getBaseConfig = function(multisite) {
             globalRoot: 'http://global.localhost:8080'
         },
 
+        //Sets the base template service reprocessing property
+        TemplateService: { useReprocess: false },
```
brianhyder: There is already a "templates" section.

include/config.js
```diff
@@ -131,6 +131,18 @@ Configuration.getBaseConfig = function(multisite) {
...
+
+        //provides configuration options for localization service
+        "localization": {
+            //enables/disables localization service using the database
+            "db":false,
+            //enables/disables localization directives being resolved in the nav
+            "nav":false,
+            //enables/disables localization directives being resolved in the content pages
+            "pages": false
+        },
```
brianhyder: There is already a "localization" section in the config.js file

include/service/entities/template_service.js
```diff
@@ -110,7 +110,12 @@ module.exports = function(pb) {
          * @property reprocess
          * @type {Boolean}
          */
-        this.reprocess = false;
+        if(pb.config.TemplateService.useReprocess){
+            this.reprocess = true;
+        }
+        else{
+            this.reprocess = false;
+        }
```
brianhyder: The config setting makes me weary.  We've seen where reprocessing everything can degrade performance.  The majority of the time the values do not need to be reprocessed.

plugins/pencilblue/controllers/actions/admin/sites/localization.js
```diff
@@ -0,0 +1,268 @@
...
+module.exports = function(pb) {
+
+    //pb dependencies
+    var util = pb.util;
+
+    /**
+     * Saves the site's Localization settings
+     */
+    function Localization(){}
+    util.inherits(Localization, pb.BaseAdminController);
```
brianhyder: There is already a localization API controller (controllers/api/localization_controller.js).  All API based localization interactions should reside there.

plugins/pencilblue/controllers/actions/admin/sites/localization.js
```diff
@@ -0,0 +1,268 @@
...
+
+    Localization.prototype.render = function(cb) {
+        var self = this;
+
+        this.getJSONPostParams(function(err, post) {
```
brianhyder: This function is deprecated.  It should be replaced by specifying the allowed request body types in the route declaration.  ```request_body: ['application/json']```.

Then inside of your controller you can get the parsed body by calling: ```this.body``` inside of your controller.  See ```getTextLengthByPostParam``` in *text_api.js* from the sample plugin as an example.

plugins/pencilblue/controllers/actions/admin/sites/localization.js
```diff
@@ -0,0 +1,268 @@
...
+            }
+
+            Localization.prototype.saveLocalesToFile(post, cb);
+        });
+
+    };
+
+    Localization.prototype.saveLocalesToFile = function(post, cb){
+        var self = this;
+        self.query = post;
```
brianhyder: instead of modifying the query params why not just pass the object that you want to process to getLocalesFromFile?

plugins/pencilblue/controllers/actions/admin/sites/localization.js
```diff
@@ -0,0 +1,268 @@
...
+        self.getLocalesFromFile(function(err, pluginsJsonFileObj) {
+            if (err)
+                throw err;
```
brianhyder: Please use braces for all if statements.  Event the one liners.  Code climate pitches a fit when we do.

include/config.js
```diff
@@ -131,6 +131,18 @@ Configuration.getBaseConfig = function(multisite) {
             globalRoot: 'http://global.localhost:8080'
         },
 
+        //Sets the base template service reprocessing property
+        TemplateService: { useReprocess: false },
+
+        //provides configuration options for localization service
+        "localization": {
+            //enables/disables localization service using the database
+            "db":false,
```
brianhyder: some managed environments are not going to allow write access to disk for running web processes.  Setting the default to false could cause errors when attempting to save overrides.  Taking the approach where all overrides are saved in the DB but are synced and held in memory by each process would eliminate the need for this setting and protect our users from disk write errors.

plugins/pencilblue/controllers/actions/admin/sites/localization.js
```diff
@@ -0,0 +1,268 @@
...
+                        code: 500,
+                        content: pb.BaseController.apiResponse(pb.BaseController.API_FAILURE, self.ls.get('ERROR_SAVING'), e)
+                    });
+                }
+                self.sendLocalesToFile(pluginsJsonFileObj, cb);
+            }
+        });
+    };
+
+    Localization.prototype.saveLocalesToDatabase = function(post, cb){
```
brianhyder: There is a lot of heavy lifting done by this controller.  abstracting it out to a service will make the controller more readable and reduce overall complexity.

The service should inherit from BaseObjectService.  It will eliminate most of the logic in this function

plugins/pencilblue/include/routes.js
```diff
@@ -60,6 +60,34 @@ module.exports = function Routes(pb){
...
+            path: '/admin/localization',
+            auth_required: true,
+            access_level: pb.SecurityService.ACCESS_EDITOR,
+            inactive_site_access: true,
+            controller: path.join(pb.config.docRoot, 'plugins', 'pencilblue', 'controllers', 'admin','sites','localizations.js'),
+            content_type: 'text/html'
+        },
+        {
+            method: 'post',
+            path: '/actions/admin/localization',
```
brianhyder: We'd like this to be: ```/api/localizations/overrides``` to fall in line with the other localization API end-points

plugins/pencilblue/include/routes.js
```diff
@@ -60,6 +60,34 @@ module.exports = function Routes(pb){
...
+            auth_required: true,
+            access_level: pb.SecurityService.ACCESS_EDITOR,
+            inactive_site_access: true,
+            controller: path.join(pb.config.docRoot, 'plugins', 'pencilblue', 'controllers','actions', 'admin','sites','localization.js'),
+            content_type: 'text/html'
+        },
+        {
+            method: 'get',
+            path: '/actions/admin/localization',
```
brianhyder: We'd like this to be: ```/api/localizations/overrides``` to fall in line with the other localization API end-points

brianhyder:
Hey all,  I have reviewed the PR and have put my comments inline.  I have concerns around the performance impact and maintainability of the current design.  It is a large change and without any supporting tests (that includes the ones we have neglected to write) in place the risk is too large at this time.  I'm happy to talk through it and address any questions but at this time we will not be merging this change set.

kylebechtel77:
@brianhyder I've read all your suggestions and will be working on implementing them throughout this week. I agree the design needs work, we pushed it back to get feedback on how you would like us to refactor it, and most of your suggestions we thought of as well and agree with. Ill keep you up to date on our refactor progress.

PR #954, opened 2016-03-01
Fixes #945 Analytics optimizations

### LINKED ISSUE #945: AnalyticsManager does not respect site
This is not a high priority issue because it is not impacting functionality. 

If you install the GA plugin on a site specific level, every site hits GoogleAnalytics.onRequest method. This is okay since there is error handling around not having the settings, but it clutters the logs with warnings.

I found the issue with the AnalyticsManager not respecting site and I will fix, but if there is a specific solution you want to implement, let me know soon. I will resolve it that way. :)

brianhyder:
@cchatham A second set of eyes would be immensely helpful.

### DIFF
--- a/include/service/entities/template_service.js
+++ b/include/service/entities/template_service.js
@@ -48,7 +48,7 @@ module.exports = function(pb) {
         else {
             localizationService = opts.ls;
         }
-        
+
         /**
          * @property localCallbacks
          * @type {Object}
@@ -65,7 +65,7 @@ module.exports = function(pb) {
         if (localizationService) {
             this.localizationService = localizationService;
         }
-        
+
         /**
          * @property activeTheme
          */
@@ -123,7 +123,7 @@ module.exports = function(pb) {
          * @type {String}
          */
         this.siteUid = pb.SiteService.getCurrentSite(opts.site);
-        
+
         /**
          * @property settingService
          * @type {SettingService}
@@ -151,6 +151,16 @@ module.exports = function(pb) {
 
     var TEMPLATE_PIECE_STATIC = 'static';
     var TEMPLATE_PIECE_FLAG   = 'flag';
+
+    /**
+     * Tracks the template keys that do not exist on disk.  This allows us to
+     * skip disk reads after the first request for a non-existent template
+     * @private
+     * @static
+     * @property MIS_CACHE
+     * @type {Object}
+     */
+    var TEMPLATE_MIS_CACHE = {};
 
     /**
      * A container that provides the mapping for global call backs.  These should
@@ -217,8 +227,8 @@ module.exports = function(pb) {
     };
 
     /**
-     * When a flag is encountered that is not registered with the engine the 
-     * handler is called as a fail safe.  It is expected to return a string that 
+     * When a flag is encountered that is not registered with the engine the
+     * handler is called as a fail safe.  It is expected to return a string that
      * will be put in the place of the flag.
      *
      * @method setUnregisteredFlagHandler
@@ -235,8 +245,8 @@ module.exports = function(pb) {
     };
 
     /**
-     * When a flag is encountered that is not registered with the engine the 
-     * handler is called as a fail safe unless there is a locally registered handler.  
+     * When a flag is encountered that is not registered with the engine the
+     * handler is called as a fail safe unless there is a locally registered handler.
      * It is expected to return a string that will be put in the place of the flag.
      *
      * @method setGlobalUnregisteredFlagHandler
@@ -253,8 +263,8 @@ module.exports = function(pb) {
     };
 
     /**
-     * Sets the option that when true, instructs the engine to inspect the content 
-     * provided by a flag for more flags.  This is one way of providing iterative 
+     * Sets the option that when true, instructs the engine to inspect the content
+     * provided by a flag for more flags.  This is one way of providing iterative
      * processing of items.  See the sample plugin for an example.
      * @method setReprocess
      * @param {Boolean} reprocess
@@ -262,9 +272,9 @@ module.exports = function(pb) {
     TemplateService.prototype.setReprocess = function(reprocess) {
         this.reprocess = reprocess ? true : false;
     };
-    
+
     /**
-     * Retrieves the active theme.  When not provided the service retrieves it 
+     * Retrieves the active theme.  When not provided the service retrieves it
      * from the settings service.
      * @private
      * @method _getActiveTheme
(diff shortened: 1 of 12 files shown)


include/service/entities/template_service.js
```diff
@@ -294,27 +313,46 @@ module.exports = function(pb) {
...
+
         //build set of paths to search through
-        var hintedTheme   = this.getTheme();
-        var paths         = [];
-        if (hintedTheme) {
-            paths.push(TemplateService.getCustomPath(this.getTheme(), relativePath));
+        var paths = [];
+        var themePath = null;
+        var hintedTheme = this.getTheme();
+        if (hintedTheme && !TemplateService.isTemplateBlacklisted(hintedTheme, relativePath)) {
```
brianhyder: The blacklisting was introduced because the profiler showed we spent a lot of time looking for templates in other plugins.  We did this on each request regardless of whether or not we had already tried it.  The blacklist reduces the number of times we ever look at disk for a theme and path combo.

include/service/entities/template_service.js
```diff
@@ -784,7 +825,7 @@ module.exports = function(pb) {
      * @return {string} The absolute path
      */
     TemplateService.getCustomPath = function(themeName, templateLocation){
-        return path.join(pb.config.docRoot, 'plugins', themeName, 'templates', templateLocation + '.html');
+        return CUSTOM_PATH_PREFIX + themeName + '/templates/' + templateLocation + '.html';
```
brianhyder: This was another one that showed up on the profiler.  We spent a lot of time in this function.  Turns out that ```path.join``` is ~60 times slower than just concatenating the strings.

include/service/memory_entity_service.js
```diff
@@ -80,13 +80,13 @@ module.exports = function MemoryEntityServiceModule(pb) {
         if(value == null && this.site !== GLOBAL_SITE && !this.onlyThisSite) {
             value = getGlobalValue(this, key);
         }
-        cb(null, value);
+        process.nextTick(function() { cb(null, value); });
```
brianhyder: This was added to give other lines of execution a chance to run.  We found that when using in memory caching the stack traces were getting too deep.  This just frees things up a bit to tackle the next item in the control flow queue

include/service/memory_entity_service.js
```diff
@@ -164,7 +164,9 @@ module.exports = function MemoryEntityServiceModule(pb) {
             rawValue[this.keyField]   = key;
             rawValue[this.valueField] = value;
         }
-        this.storage[this.site] = {};
+        if (util.isNullOrUndefined(this.storage[this.site])) {
+            this.storage[this.site] = {};
+        }
```
brianhyder: bug fix that only allowed one key in storage at a time

include/system/analytics_manager.js
```diff
@@ -22,14 +22,35 @@ var util   = require('../util.js');
...
+         */
         this.site = options.site || pb.SiteService.GLOBAL_SITE;
+
+        /**
+         * The amount of time, in milliseconds, that the manager will wait for
+         * each provider to render.
+         * @property timeout
+         * @type {Integer}
+         */
+        this.timeout = options.timeout || DEFAULT_TIMEOUT;
```
brianhyder: Timeouts are now configurable.  See config.js for more info: ```pb.config.analytics.timeout```

include/system/analytics_manager.js
```diff
@@ -108,11 +136,184 @@ module.exports = function AnalyticsManagerModule(pb) {
...
+            var d = domain.create();
+            d.run(function() {
+
+                //set the execution timeout.  We set it here and not outside of the
+                //domain to ensure that it gets a fair shot at running for the entire
+                //allotted timeout.  Setting it outside makes it subject to other lines
+                //of execution already in queue that will burn through the timeout
+                //before execution of this domain begins.
+                opts.start = new Date().getTime();
+                opts.th = setTimeout(AnalyticsManager.buildTimeoutHandler(opts), timeout);
```
brianhyder: Note the comment. Putting it outside of the domain was causing contention with other lines of execution essentially robbing the provider of the time allowed to execute when under load.

include/system/analytics_manager.js
```diff
@@ -108,11 +136,184 @@ module.exports = function AnalyticsManagerModule(pb) {
...
+     * Updates the runtime statistics to include
+     * @static
+     * @method getStats
+     * @return {Object}
+     */
+    AnalyticsManager.updateStats = function(startTime, provider) {
+        var runTime = (new Date().getTime()) - startTime;
+        updateStatBlock(STATS.cumulative, runTime);
+        updateStatBlock(STATS.providers[provider], runTime);
+    };
```
brianhyder: As part of performance tracking we now keep stats on the average runtime for the providers (except when it errors).  It tracks individual providers (not on a site basis) and overall in the event that there are multiple providers

include/system/analytics_manager.js
```diff
@@ -181,30 +399,48 @@ module.exports = function AnalyticsManagerModule(pb) {
...
+        var managerInstance = new AnalyticsManager(context);
         managerInstance.gatherData(req, session, ls, cb);
     };
 
-    function checkForHook(site) {
-        return !PROVIDER_HOOKS[pb.SiteService.GLOBAL_SITE] ||
-            Object.keys(PROVIDER_HOOKS[pb.SiteService.GLOBAL_SITE]).length === 0
-        !PROVIDER_HOOKS[site] ||
-        Object.keys(PROVIDER_HOOKS[site]).length === 0;
-    }
```
brianhyder: This function was removed because it was redundant to getKeys because it had to iterate over the hash to create the key array.  Instead, I took the approach of attempting to get the keys then decide based on the length since it had to call ```Object.keys``` anyway.

include/system/analytics_manager.js
```diff
@@ -181,30 +399,48 @@ module.exports = function AnalyticsManagerModule(pb) {
...
-        var keyArray = Object.keys(PROVIDER_HOOKS[pb.SiteService.GLOBAL_SITE]) || [];
-        if(PROVIDER_HOOKS[site] && Object.keys(PROVIDER_HOOKS[site]).length > 0) {
-            return keyArray.concat(Object.keys(PROVIDER_HOOKS[site]))
+        //check for non-global site keys
+        if(PROVIDER_HOOKS[site] && site !== pb.SiteService.GLOBAL_SITE) {
+            Object.keys(PROVIDER_HOOKS[site]).forEach(keyFunc);
         }
-        return keyArray;
+        return Object.keys(keyHash);
     }
```
brianhyder: There was a small bug when in single site mode.  It would call a provider twice because it did not verify that the provider had already been added if it belonged to the global site too.  A hash is built of the provider keys from both then the distinct union of the two is used as the array of keys that is returned.

package.json
```diff
@@ -53,7 +53,7 @@
   },
   "main": "./pencilblue.js",
   "engines": {
-    "node": ">= 0.11.0"
+    "node": ">= 0.11.0 <5"
```
brianhyder: FYI... a unit test broke with node v5.  Until we can investigate further I've updated the docs and the package.json file to ensure that users are running with proper versions.

brianhyder:
@blakecallens FYI...

PR #965, opened 2016-03-16
fixes #955 - generated URLs are safe

### LINKED ISSUE #955: Url generator generates invalid URL when using colon in article title
Hi, 

When the title of an article includes a colon, the generated URL does not delete the colon.
For exemple an article title such as "Pencilblue : A CMS" would generate automatically "pencilblue-:-a-cms" as the article URL, which is invalid and leads to a 404 when we click on "Read more".

I am trying to figure out which file handles this generation to correct that.

blakecallens:
Created an external template for the generateUrl function because the ^ of the regex was breaking the template service, and because it's duplicate code anyway.

### DIFF
--- a/plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
+++ b/plugins/pencilblue/templates/angular/admin/content/articles/article_form.html
@@ -168,11 +168,7 @@
 			});
 		};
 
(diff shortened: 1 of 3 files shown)


brianhyder:
Looks good to me.  Merge when ready.

PR #966, opened 2016-03-16
Adds comment content validation

n1ru4l:
Comments could be saved into the database with empty content.
Now the new_comment api controller will check for a min-length of 5 characters.


brianhyder:
Thanks for the PR.  I will be merging it in.  

Just as an FYI... The comment controllers are deprecated and will be ripped out soon.  There is another comment API that has already been implemented and will most likely be more useful if you are integrating it into your own plugins.

https://github.com/pencilblue/pencilblue/blob/master/plugins/pencilblue/include/routes.js#L1250

PR #967, opened 2016-03-17
This changes the behavior of redirects

kylebechtel77:
instead of simply showing you the same site (even though your url is different) we now do a true redirect to the correct site.

### DIFF
--- a/include/http/request_handler.js
+++ b/include/http/request_handler.js
@@ -767,7 +767,7 @@ module.exports = function RequestHandlerModule(pb) {
 
         // If we need to redirect to a different host
         if (!siteObj && redirectHost && RequestHandler.sites[redirectHost]) {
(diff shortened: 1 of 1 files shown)


include/http/request_handler.js
```diff
@@ -765,9 +765,9 @@ module.exports = function RequestHandlerModule(pb) {
         var siteObj = RequestHandler.sites[hostname];
         var redirectHost = RequestHandler.redirectHosts[hostname];
 
-        // If we need to grab a host redirect
+        // If we need to redirect to a different host
         if (!siteObj && redirectHost && RequestHandler.sites[redirectHost]) {
-            siteObj = RequestHandler.sites[redirectHost];
+            return this.doRedirect(pb.SiteService.getHostWithProtocol(redirectHost), '301');
```
brianhyder: Instead of hard coding the ```301``` please use the constant through the status codes package.

https://www.npmjs.com/package/http-status-codes

PR #995, opened 2016-03-31
Global Custom Object Types

lgenet:
Custom Object tweaks as discussed.  This allows for types to exist at the global scope, and be instanced on a site by site basis.  Mostly just changing onlyThisSite in the Custom Object Service to false instead of true, also required loading and merging of the global and site list for the manage type page.


lgenet:
Switching to 0.6.0

PR #999, opened 2016-04-04
Add loc to admin panel

kylebechtel77:
This PR adds a side nav item to the admin panel that allows you to edit and override localizations via the current file based system. It also makes localizations site and active theme aware.

This is the order of localization selection: 
1. Site/active theme
2. Site/any theme
3. Active theme
4. Any theme
5. Generics

It also adds a few configuration options:
localization.db = true will set the localization file to use a database instead of files
localization.nav = true will let you put localizations directives in your navigation items
localization.pages = true will let you put localization directives in content pages. 

Finally: theres a caveat to the way this is being done. If you are using files instead of a db you will be able to change localizations in the admin panel, but if you are deploying to a server you run the risk of overwriting those file changes when you push. We would strongly encourage using the db if you want to edit localizations on the fly in the admin panel.


lgenet:
@brianhyder @blakecallens Hey, just checking in.  Do we have any status updates on this PR, is there anything you need from Kyle or I?  Just checking. 

-Logan

blakecallens:
@lgenet @kylebechtel77 we were blocked from merging to prevent stepping on the toes of the localization refactor in 0.7.0.

With that release out yesterday, @brianhyder and I are still debating whether a feature like this needs to be part of the core platform or if it could be accomplished with a plugin. We want to make sure that the only things we add to the core platform are either something that almost all users will need or something that a reasonable amount of users will need and that can't be done through plugins.

We definitely think it's a cool feature, we're just debating the format in which it should be available. Let's start a side conversation between the four of us on that.

kylebechtel77:
Closing this after conversations with Blake and Brian. This will need to be pulled out into either middleware or a separate plugin.

PR #1001, opened 2016-04-04
Fixes #998 - Upgrades Node JS Dependencies

### LINKED ISSUE #998: Dependency Updates
The main dependencies will be fixed in #972.  We will also address dev dependencies as an add on tracked through this issue.

brianhyder:


brianhyder:
@chrisparnin I wanted to give you a heads up that after applying your patch and testing a few things I think we have everything updated all the way with the exception of node mailer.  That will have to come later as a feature due to the amount of work necessary to upgrade.

Thank you for the work you put into this.  It made our testing efforts easier and we are more confident that the system will be stable.  If there are no objections I will close the other pull request.

PR #1071, opened 2016-06-08
Feature/1006

brianhyder:
Improves maintainability of PluginService(s)

**Description:**
Breaks out the localization loading, service initialization from the PluginService into maintainable services that are abstracted out so that other implementations can be created.

@pencilblue/owners


brianhyder:
wrong freaking base branch

PR #1085, opened 2016-07-11
Session check

lgenet:
This was the one fix that needed to be committed back to the platform that we found to be crashing the server.  There was a use case where session was not being returned from pb.session.open  We are not sure exactly how a user was managing to provide a request that would not allow this function to properly generate a session, however it was not being checked and was causing some errors for the server.  

Just committing back this fix, you might have more insight into why this might have been an issue or a more graceful/lower level change to fix this same issue.

### DIFF
--- a/include/http/request_handler.js
+++ b/include/http/request_handler.js
@@ -549,7 +549,9 @@ module.exports = function RequestHandlerModule(pb) {
             if (util.isError(err)) {
                 return self.serveError(err);
             }
-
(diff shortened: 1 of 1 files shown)


brianhyder:
@lgenet what implementation of the session store were you using? Mongo or Redis?

lgenet:
@brianhyder We are using Redis. 
 ```    "session": {
        "storage": "redis"
    },```

Issue #884, opened 2015-12-01
Localization Structure Simplification


brianhyder:
Q&A between @cchatham and @brianhyder as started in issue #501:

> Localization supports active theme but it is not leveraged by the RequestHandler. Therefore active theme is not respected. Was this intended?

No, this was oversight on my part and I will circle back around on this.

> Could we get some explanation/documentation on the Localization.storage object? It seems more nested and complex than before. We have some ideas about how to simplify it but you probably have a reason for the design we are unaware of.

It is a bit more complex than it needs to be.  Specifically, instead of having flattened keys to represent the way we reference them i.e. "general.LOCALE_DISPLAY" I built it such that LOCALE_DISPLAY key would be inside of a "general" object.  However, once the key is located the structure is optimized to provide the greatest number of overrides from a specific country, plugin, or language perspective.  The structure will also allow us to eventually provide site specific overrides if desired.  

The purpose will be to provide:

* adaquate documentation
* simplify the storage structure
* set the active theme and ensure it is leveraged

PR #1087, opened 2016-07-11
Multilingual: Force Locale in Route

lgenet:
New Feature

**Description:**
Allows for the site to do a redirect from /, or any route, to /en-US (or whatever the current locale is.)  It is toggle-able via the admin site panel.  This will force your site to use the locale in the url no matter how you hit the site or what route you hit.  If you hit /en-US then it will NOT do the redirect since you already have locale in your url.

This will pull from the cookie (the locale or it will use the sites default locale if the locale does not exist.)

@pencilblue/owners


include/http/request_handler.js
```diff
@@ -172,6 +172,7 @@ module.exports = function RequestHandlerModule(pb) {
           active: site.active,
           uid: site.uid,
           displayName: site.displayName,
+          forceLocale: site.forceLocale,
```
brianhyder: Please update the documentation for the method signature:
```
/**
 * @static
 * @method loadSite
 * @param {Object} site
 * @param {boolean} site.active 
 * @param {string} site.uid 
 * @param {string} site.displayName 
 * @param {boolean} site.forceLocale 
 * @param {string} site.hostname 
 * @param {string} site.defaultLocale 
 * @param {Array} site.supportedLocales 
 * @param {Array} site.prevHostnames
 */
```

include/http/request_handler.js
```diff
@@ -534,15 +535,26 @@ module.exports = function RequestHandlerModule(pb) {
...
 
+        if(typeof(siteObj) !== 'undefined' && siteObj.hasOwnProperty('forceLocale') && siteObj.forceLocale) {
+            cookies.locale = cookies.locale || cookies[" locale"];
+            cookies.locale = cookies.locale || siteObj.defaultLocale;
+            if (checkRouteContainsLocale(url, siteObj)) {
+                var redirectLocation = pb.SiteService.getHostWithProtocol(this.hostname);
+                redirectLocation += '/' + cookies.locale + this.url.path;
+                return this.doRedirect(redirectLocation, pb.HttpStatus.MOVED_TEMPORARILY);
+            }
+        }
```
brianhyder: I think this is the wrong spot to do this.  If you wait until the route is derived then you can check the route spec to see if the route is localized instead of having to guess through your ```checkRouteContainsLocale```.  Your implementation assumes routes are localized if a supported locale exists within URL path.  This is an incorrect assumption.

include/service/jobs/sites/site_create_edit_job.js
```diff
@@ -99,6 +99,7 @@ module.exports = function SiteCreateEditJobModule(pb) {
                     site.displayName = mySite.displayName || site.displayName;
                     site.supportedLocales = mySite.supportedLocales || site.supportedLocales;
                     site.defaultLocale = mySite.defaultLocale || site.defaultLocale;
+                    site.forceLocale = (mySite.forceLocale!==null) ? mySite.forceLocale : site.forceLocale;
```
brianhyder: ```util.isNullOrUndefined(mySite.forceLocale) ?  site.forceLocale: mySite.forceLocale;```

include/http/request_handler.js
brianhyder: The other piece I'm confused about is how a locale cookie gets set.  This is not a platform concept.  It would be encouraged to put your preferred locale for a user in the session object.

include/http/request_handler.js
```diff
@@ -564,6 +576,25 @@ module.exports = function RequestHandlerModule(pb) {
...
+            var loca‌‌les =  Object.keys(siteObj.supportedLocales);
+            for (var i = 0; i < loca‌‌les.length; i++) {
+                if(url.indexOf(loca‌‌les[i]) != -1){
+                    return false;
+                }
+            }
+        }
+
+        return true;
+    }
```
brianhyder: As I mentioned above, I don't think this function is necessary once the route is derived.  You will be able to check the localized property

include/http/request_handler.js
lgenet: On where the cookie gets set.  This is going to be the hard part for me.  We have a base plugin that contains an angular controller that is included on every page (in the header) and it will set the cookie by getting the locale that we have derived for the user.  Since this was a front end piece to be loaded by the html and did not live at the platform level, I was wondering how best it would be to submit that back to you all.

lgenet: Do you want this moved inside of the derive localization function?  Or do you want us to move it to after it is called which will be in a few different places.  I do think it will be better to use that instead of guessing at the locale.

brianhyder: I don't think i have a good handle on what the angular plugin is supposed to do?  Is it forcing a specific locale other than what is being sent in the locale header? We parse the locale header each time it is sent to the server (seems to be exactly what the cookie is doing).   

We won't accept this PR until we get a better understanding of why this is needed in the platform and can ensure that all pieces of the puzzle cookie creation and value extraction are part of the platform.  It doesn't make sense for us to put the other half of a plugin into the platform.

Is the real issue that you can't hook into the request handler from a plugin?

plugins/pencilblue/templates/admin/sites/site_form.html
```diff
@@ -35,6 +35,13 @@
                         </select>
                         <div class="form_error" ng-if="defaultLocale==''">^loc_REQUIRED_FIELD^</div>
                     </div>
+                    <div class="form-group"> 
+                        <label>Force Locale</label>
```
brianhyder: Please add localizations for this key

plugins/pencilblue/templates/admin/sites/site_form.html
```diff
@@ -35,6 +35,13 @@
                         </select>
                         <div class="form_error" ng-if="defaultLocale==''">^loc_REQUIRED_FIELD^</div>
                     </div>
+                    <div class="form-group"> 
+                        <label>Force Locale</label> 
+                        <div class="btn-group"> 
+                            <button type="button" class="btn btn-default" ng-class="{'active': forceLocale}" ng-click="forceLocale = true">^loc_YES^</button>
```
brianhyder: Coming in #884 is a deprecation warning for localization flags that aren't fully qualified.  Please update these to 
```
^loc_generic.YES^
^loc_generic.NO^
```

include/http/request_handler.js
lgenet: done is done :) I added the forceLocale to the comment.

include/http/request_handler.js
lgenet: The ultimate goal of this code and our goal with this set is to force the locale into the route.  In the Multilingual work that has already been done, you added support for locale in the route if you specify the right prop int he route object.  (Localization: true)

That _allows_ the user to use / the same as /en-US or /fr-FR or ect.  Our use case was to say if the user types in: www.mysite.com, we need to redirect to www.mysite.com/en-US  We wanted to have a way to force the user to get the locale in the route, regardless of whether or not they typed it in.  

We did this by storing their locale in a cookie.  The code in the request handler was to read that cookie we set, and then redirect the user from / to /en-US or /fr-FR or whatever locale the cookie says they should be on. 

I would not be opposed to handling this in a plugin, if there was a way to hook into the request handler from a plugin.  I also would not mind if there was some way to build in this ability (to force the locale into the route) at the platform level.

I hope that makes more sense, if you have any further questions please let me know.

brianhyder:
OK... I looked back over this again.  The biggest problem that y'all are running into is that the RequestHandler series of execution does not provide any hooks.  The ideal solution would be to intercept the request just after the themed version of the route is derived.  This would allow for redirects or what ever else you would need to do to modify the request, session, etc.  The second piece would be setting the locale preference.  I would set this in the session through an API endpoint rather than trying to set a cookie.  This way it is session based and can be retrieved cross domain if needed.  

I have created an experimental branch off of master (feature/reqhandlerhooks) that creates a hook in the request handler so that you can perform your redirect or request/response modification after the route has been derived.  Please take a look when you get a chance, specifically: https://github.com/pencilblue/pencilblue/commit/74b02f9149d3f976046f8349b72c71f39d80eaff

I would pull down the branch to see if that implementation would work for you guys.  If it does let me know and I will make the branch production ready.

lgenet:
@brianhyder It might be this weekend before I get a chance to pull the branch and test it out.  Thank you so much for you help and getting a working prototype up so quickly!

lgenet:
@brianhyder What do you mean by routeSpec?  I do not see an object on reqHandler with that name.  What is the proper thing we need to access to get locale in the onStartupWithContext method?

brianhyder:
I believe I was referring to the themeRoute property on the context object that is provided during the callback. See here:
https://github.com/pencilblue/pencilblue/commit/74b02f9149d3f976046f8349b72c71f39d80eaff#diff-4ca7642169f8885ed955346252fff3cdR1010

So first step would be to register your handler in your main module file:
```
SamplePlugin.onStartupWithContext = function (context, cb) {
  
         pb.RequestHandler.on(pb.RequestHandler.THEME_ROUTE_RETIEVED, function(ctx, cb) {
             //here the ctx object has two properties, ```themeRoute``` and ```requestHandler```. You can console.log it to see what properties it has available.  If you need more data we can expand the context or get it directly from the RequestHandler instance.
         });
}
```

lgenet:
@brianhyder Also wanted to let you know that after @kylebechtel77 and I looked into it, we believe that your event emitter function will give us the control and flow that we need.  So that would be super awesome to get that merged in and production ready.  Thank you very much.

brianhyder:
Awesome. I'll see if I can't get that merged back in. Hoping to have a lull
moment over the holiday weekend.

PR #1088, opened 2016-07-11
Template Service Get

lgenet:
New Feature

**Description:**
This provides a function to get the values that were registered via ts.registerLocal()  This allows for a value to be registered in one service, and then verified that it was registered in another service.

Use Case:  We registered a local in our controller.  We then called to a service to do some processing. We then wanted to see if that local was registered, if it was we would do one code flow, if not we would do another.

@pencilblue/owners

### DIFF
--- a/include/service/entities/template_service.js
+++ b/include/service/entities/template_service.js
@@ -236,10 +236,10 @@ module.exports = function(pb) {
     /**
      * returns the value associated with a registered local key(flag)
      *
-     * @method registerLocal
+     * @method getRegisteredLocal
      * @param {string} flag The flag name to map to the value when encountered in a
      * template.
-     * @return {callbackFunctionOrValue} the callback or the value that was assigned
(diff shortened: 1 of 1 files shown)


include/service/entities/template_service.js
```diff
@@ -234,6 +234,20 @@ module.exports = function(pb) {
     };
 
     /**
+     * returns the value associated with a registered local key(flag)
+     *
+     * @method registerLocal
```
brianhyder: please change this getRegisteredLocal

include/service/entities/template_service.js
```diff
@@ -234,6 +234,20 @@ module.exports = function(pb) {
...
+     * @param {string} flag The flag name to map to the value when encountered in a
+     * template.
+     * @return {callbackFunctionOrValue} the callback or the value that was assigned
```
brianhyder: please change to: ```{*}```

PR #1092, opened 2016-07-12
Czech translation

tomasvarg:
The Czech language localization


brianhyder:
This is awesome, thank you!

PR #1101, opened 2016-07-19
Feature/977

### LINKED ISSUE #977: Custom Sanatization Filter
I need to create pages that have links to named anchors in the WYSIWYG editor (so the user can later edit them). I could make this a custom page theme, but that will not allow the user to edit page. 
What is the best way to create a sanitization filter to omit the name divs for anchor links?

Thanks!

Forget the one by Momentage, I should not have been logged in as their account when I sent that.

brianhyder:
Fixes #977 

**Description:**
Adds the ability to override the sanitization rules for content fields as well as the generic fields that strip all HTML formatting.

@pencilblue/owners


brianhyder:
Wrong branch.... FML

PR #1120, opened 2016-08-03
Allow for grouping of plugin settings

lgenet:
Fixes #

- [ ] Fix Bug Line 96
- [ ] Revise Insert Group Headings
- [ ] Make Functions Public
- [ ] Update line 105

**Description:**
This is a little bit of an odd/different PR than the ones I have been doing far more recently.  Let me start off by explaining our use case and then move into the 'defaulting' measures put in place.

**Use Case:**  In my use case for creating plugins, we have a specific one with upwards of forty settings or so.  The current Settings page does not have ids on the elements.  Thus when we wrote our selenium (integration tests) over some of our features we had to use child selectors to manipulate settings.  Thus having IDs would be useful.

Beyond that, with having forty some odd settings, it becomes easy to misplace them.  Also with having so many, there tend to be patterns.  Let say five settings are related to one feature, another three are related to another feature and so on.  It would be helpful to group all of the like settings together with a handy title to help navigation through the vast number of settings.

**Defaulting measures:** So with adding these changes, one might be afraid of what would happen if you dont specify a header.  Code has been added to set it to 'default' if you do not specify one in your details.json file.  We also sort the settings alphabetically and then sort the groupings alphabetically.  To prevent showing the default title, we also make sure that they appear at the top of the list before all the other setting groups.  We do not show their title since default was an arbitrary choice and not something the user would necessarily want to use.

So bottom line, if you dont use this feature you will not see any change from the way plugin settings currently work. 

@pencilblue/owners


plugins/pencilblue/controllers/admin/plugins/plugin_settings.js
```diff
@@ -97,6 +98,12 @@ module.exports = function(pb) {
                     }
```
brianhyder: Doh!!! Can you fix my bug on line 96?  Should be ```else if (!isNaN(item.value)) {```

plugins/pencilblue/controllers/admin/plugins/plugin_settings.js
```diff
@@ -128,6 +135,36 @@ module.exports = function(pb) {
...
+        var previous = clone[clone.length - 1], current;
+        for (var i = clone.length - 1; i >= 0; i--) {
+            current = clone[i];
+            if (current.heading != previous.heading) {
+                clone.splice(i+1, 0, {id: previous.heading, isHeading: true, heading: convertNameToDisplayName(previous.heading)});
+            }
+
+            previous = current;
+        }
+    }
```
brianhyder: This function doesn't seem to handle the case of displaying a header if the first setting has a group.  I'd like to propose the following:
```
    /**
     * Inserts a new record before the start of the new group.  This assumes that the settings array is already sorted 
     * by group
     * @param {Array} settings
     * @returns {Array}
     */
    PluginSettingsFormController.insertGroupHeadings = function(settings) {

        var prevHeading = '';
        var settingsAndHeaders = [];
        for (var i = 0; i < settings.length; i++) {
            if (prevHeading !== settings[i].heading) {
                settingsAndHeaders.push({
                    id: settings[i].heading,
                    isHeading: true,
                    heading: convertNameToDisplayName(settings[i].heading)
                });
            }
            prevHeading = settings[i].heading;
            settingsAndHeaders.push(settings[i]);
        }
        return settingsAndHeaders;
    };
```

plugins/pencilblue/controllers/admin/plugins/plugin_settings.js
```diff
@@ -128,6 +135,36 @@ module.exports = function(pb) {
...
+    function sortOn(property){
```
brianhyder: Starting in 0.7.0 we are going to bump the minimum node version to 4.x.  We are doing this in order to provide support for ES6 classes.  We'll slowly be making that transition over the course of the next year or so.  In support of that we ask that all functions are declared as part of the prototype, ```PluginSettingsFormController.sortOn = function(property) {```.  This will make it easier for us to convert as well as keep functions public so that they can be unit tested.

plugins/pencilblue/controllers/admin/plugins/plugin_settings.js
```diff
@@ -97,6 +98,12 @@ module.exports = function(pb) {
                     }
                 }
 
+                // First sort the settings alphabetically, then by heading/group
+                clone.sort(sortOn("displayName"));
+                clone.sort(sortOn("heading"));
+
+                insertGroupHeadings(clone);
```
brianhyder: Based on the recommended changes for this function this line will need to update to:
```clone = PluginSettingsFormController.insertGroupHeadings(clone);```

brianhyder:
This makes total sense and a great idea.  I'm happy to merge things in but I have a couple of requests:

* Update to merge into 0.7.0.  For new features we like to have go out with the next release.  It keeps us all out of merge hell when big sweeping changes come through since we typically do code smell cleanup for each new release.
* From the controller down to the details.json file I'd like change from "header" to "group".  This, IMO, will make more sense for the majority of the platform since that is how they are categorized.  
* Allow for an explicit *displayName* property for each setting.  If not provided, fall back to the function that converts the *name* property to the display name.
* Add validations for each new setting property: *displayName*, *group*.  Each can be optional but must be strings.

I understand this bumps the scope up a bit but it really will be a good addition to the platform.  Let me know if you need any help.  I should have time this week to lend a hand.

plugins/pencilblue/controllers/admin/plugins/plugin_settings.js
kylebechtel77: Fixed :)

PR #1126, opened 2016-08-08
fixes #1109 - reset password in angular

### LINKED ISSUE #1109: more about forgot password issue
Sorry, I was telling you that everything was working properly but not completely. The email with the link leads us to the change password view, but here, current password is required fields. Obviously I can not go ahead, because I do not remember my password. As a result I can not save new password and then functionality does not reach the end.

pencilblue: v0.6.0",
nodeJs:  v5.10.1
mongodb: v3.2.4

blakecallens:
Fixes #1109 

- [X] Unit tests created and pass

**Description:**

reset_password session variable pushed to Angular

@pencilblue/owners

### DIFF
--- a/plugins/pencilblue/controllers/user/change_password.js
+++ b/plugins/pencilblue/controllers/user/change_password.js
@@ -61,10 +61,11 @@ module.exports = function (pb) {
                 return self.redirect('/', cb);
             }
 
-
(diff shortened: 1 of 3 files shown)


brianhyder:
This doesn't appear to change the behavior of the page after the password change.  After a password reset I would expect that I am redirected to the login page.  Additionally, this view still shows the menu options (change password, manage account).  I would expect the reset password view to exclude menu options for managing an account and only provide the ability to change the password.

thoughts?

PR #1134, opened 2016-09-04
Feature/1131

### LINKED ISSUE #1131: Extract Plugin Initialization from PluginService
**Description:**
The plugin service is bloated (2k+ lines) with items that are beyond its core responsibility.  The plugin initialization should be extracted such that it is contained within its own service.  This will make it testable, maintainable and extendable.

**Steps to Reproduce:**
n/a

**Configuration Details:**
n/a

**Log Output:**
n/a

PencilBlue Version: 0.7.0

Node.js Version: n/a

Environment Description: n/a

brianhyder:
Fixes #1131 

**Description:**
Abstracts the Plugin Initialization from the plugin service into its own testable service.

@pencilblue/owners

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-  - "0.11"
-  - "0.12"
   - "4"
   - "5"
   - "6"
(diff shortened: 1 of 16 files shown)


brianhyder:
@kylebechtel77 @lgenet Just a heads up on this.  I wanted to give y'all an opportunity to take a look and test before it got merged.  The goal was to abstract out so everything was testable but I think we gained a bit of performance in the process.

PR #1135, opened 2016-09-05
Request Handler Hook - On Theme Route Retrieved

brianhyder:
Fixes #

**Description:**
Provides a framework so that events can be emitted and handled from the RequestHandler.  This allows for better control flow in specific conditions where redirects need to occur

@pencilblue/owners
@lgenet @kylebechtel77

### DIFF
--- a/plugins/sample/sample.js
+++ b/plugins/sample/sample.js
@@ -72,9 +72,8 @@ module.exports = function SamplePluginModule(pb) {
     SamplePlugin.onStartupWithContext = function (context, cb) {
 
         pb.RequestHandler.on(pb.RequestHandler.THEME_ROUTE_RETIEVED, function(ctx, cb) {
-            if (ctx.themeRoute.theme === 'pencilblue') {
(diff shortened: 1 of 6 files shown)


plugins/sample/sample.js
```diff
@@ -72,6 +72,21 @@ module.exports = function SamplePluginModule(pb) {
...
+         * @param ctx {object}
+         * @param {RequestHandler} ctx.requestHandler
+         * @param {object} ctx.themeRoute
+         * @param {function} (Error)
+         */
+        pb.RequestHandler.on(pb.RequestHandler.THEME_ROUTE_RETIEVED, function (ctx, callback) {
+            //do what ever needs to be done.  Use the callback to continue normal control flow or don't if you need to do redirects
+            pb.log.debug('SamplePlugin: The request handler hook triggered for request: %s', ctx.requestHandler.url.path);
+            callback();
+        });
```
brianhyder: Please make sure to read this part

PR #1147, opened 2016-09-27
Navigation nesting

webdesus:
Hi everybody.
I the recently start work with pencilblue and for me for my project very not enough nested navigation.
Therefore I began work on it, but after yesterday publish new release pencilblue decided send what I have now, whatever you do double work.

This commit includes:
**1)** Removed error in "navigation map" associated with the pressing beside with element name slightly to the right will handle action delete element.
**2)** I did feature nested navigation. Now container might put inside another container to infinity.
**3)** Container name might used in path when set option "use in a path". If no conflict in paths.
	For example we have next navigation map(for better understanding in example use fields from 2 collections "navigation map" and "sections"):
```
	{
		name : 'level_1',
		type : 'container',
		use_in_path : true
		children [
			{
				name : 'level_2',
				type : 'container'
				children :[
					{
						name : 'level_3',
						type : 'container',
						use_in_path : true
						children : [
							{
								name : 'level_3_1',
								type : 'page',
								url : 'page/test1'

							},
							{
								name : 'level_3_2',
								type : 'page',
								url : 'page/test2'


							}
						]

					},
					{
						name : 'level_2_1',
						type : 'page',
						url : 'page/test3'
					},
					{
						name : 'level_2_2',
						type : 'page',
						url : 'page/test4'

					}

				]


			},
			{
				name : 'level_1_1'
				type : 'page',
				url : 'page/test5'
			}
							
		]
	}
```
**Now we have opportunity using alternative paths using navigation path:**
```
	page/test1 -> level_1/level_3/level_3_1
	page/test2 -> level_1/level_3/level_3_2
	page/test3 -> level_1/level_2_1
	page/test4 -> level_1/level_2_2
	page/test5 -> level_1/level_1_1

```

**4)** In inside section element I deleted field "parent". I think user occurs misunderstanding work navigation map. When he in section set parent, then change position on navigation map and after decided changed name, description etc inside section, but not parent of section, he get not correct navigation map. Yes we can reset parent in all sections while save navigation map, but whether it is necessary?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ config.json
 config.js
 
 npm-debug.log
+
+/node_modules
--- a/include/http/request_handler.js
+++ b/include/http/request_handler.js
@@ -105,6 +105,8 @@ module.exports = function RequestHandlerModule(pb) {
     RequestHandler.index   = {};
     RequestHandler.sites = {};
     RequestHandler.redirectHosts = {};
+    RequestHandler.sectionsPaths = {};
+
     var GLOBAL_SITE = pb.SiteService.GLOBAL_SITE;
     /**
      * The internal storage of static routes after they are validated and processed.
@@ -152,6 +154,7 @@ module.exports = function RequestHandlerModule(pb) {
             var result;
             try {
                 result = RequestHandler.registerRoute(descriptor, RequestHandler.DEFAULT_THEME);
+                pb.SectionService.updateSectionsPaths();
             }
             catch(e) {}
             if (!result) {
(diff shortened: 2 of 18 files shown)


blakecallens:
Thanks, @webdesus. We released 0.7.0 yesterday, so you'll need to sync your pull request up to master. Then we can take a look at it.

webdesus:
@blakecallens I did it. I very hurried and not write tests (your tests work fine on my pc). But if this commit will interesting for you I can make it.

webdesus:
I created separate field for navigation path, because user can create not valid name for path(name might include not valid symbols or language). And now page without container too might participate in path.

PR #1149, opened 2016-10-01
Find media by topic

webdesus:
Fixed "search_input.html". Event "clearSearch"  does not always work.
And makes opportunity find media by topic. I have many images separated by types using topics. It's like folder in OS.


plugins/pencilblue/templates/angular/admin/content/media/manage_media.html
```diff
@@ -33,8 +33,25 @@
 			sortDesc: false
 		}];
 
+		$scope.setItem = function(item) {
+				$scope.topic = item;
+				$scope.search();
+				$scope.$apply();
```
blakecallens: Shouldn't need a $scope apply for this to work. Two way data binding should cover it.

public/localization/en-US.js
```diff
@@ -295,7 +295,8 @@ module.exports = {
         TEMPLATE: 'Template',
         HEADLINE: 'Headline',
         SUBHEADING: 'Subheading',
-        PUBLISH_DATE: 'Publish date'
+        PUBLISH_DATE: 'Publish date',
+        PAGE_NAME: 'Page name'
```
blakecallens: This will need to be added to all localization files.

plugins/pencilblue/controllers/admin/content/media/manage_media.js
```diff
@@ -57,6 +58,10 @@ module.exports = function(pb) {
                 var title = self.ls.g('media.MANAGE_MEDIA');
                 self.setPageName(title);
                 self.ts.registerLocal('angular_objects', new pb.TemplateValue(angularObjects, false));
+                self.ts.registerLocal('content_type', 'topic');
+                self.ts.registerLocal('selection_id_field', 'name');
+                self.ts.registerLocal('content_search_value', '');
```
blakecallens: Why register a local as an empty value?

plugins/pencilblue/controllers/admin/content/navigation/nav_item_form.js
```diff
@@ -63,7 +63,7 @@ module.exports = function(pb) {
             self.setPageName(self.navItem[pb.DAO.getIdField()] ? self.navItem.name : self.ls.g('generic.NEW_NAV_ITEM'));
             self.ts.registerLocal('angular_objects', new pb.TemplateValue(angularObjects, false));
             self.ts.registerLocal('content_type', '{{section.type}}');
-            self.ts.registerLocal('selection_id_field', 'item');
+            self.ts.registerLocal('selection_id_field', 'headline');
```
blakecallens: Why an edit to navigation in a pull for searching media by topic?

webdesus: I don't remember why i decided using selection_id_field for filter(now i think angular object will be enough), but this field not using in code. [Why it exists](https://github.com/pencilblue/pencilblue/blob/master/plugins/pencilblue/templates/admin/elements/content_search.html#L2)?

plugins/pencilblue/controllers/admin/content/media/manage_media.js
webdesus: `content_search.html` element have default value `content_search_value`.

public/localization/en-US.js
webdesus: I was not sure of the approval of the idea. I think if idea will approved then i add this values to all localization files.

plugins/pencilblue/templates/angular/admin/content/media/manage_media.html
webdesus: it's was copy/paste from [here](https://github.com/pencilblue/pencilblue/blob/master/plugins/pencilblue/templates/angular/admin/content/navigation/nav_item_form.html#L13). I decided will be problem with popmenu if i remove this code.

webdesus:
I think use this code like plugin bad idea. I can write this functional and elsewhere (pages, articles etc). I have some thoughts on version 1.0. And I decided share with you( @brianhyder  and @blakecallens). [link](https://docs.google.com/presentation/d/1FKU7V8WOo5-WugqGQR-D7hQFRWxnh6HvtfV_PZ2SFLE/edit?usp=sharing) on slides

webdesus:
I implemented this feature for page and articles. I think for custom objects  not needed implemented this feature, cuz object might have many fields type topics. 
And question about localization. I used to think if users localisation not have variable on him language taken english by default. But now i see many different language had english words. I should create new field on each files using google translate or copy english realization?

PR #1153, opened 2016-10-03
Fixed dao service method saveBatch

webdesus:
After version 2.6 api mongodb changed https://docs.mongodb.com/manual/reference/method/js-bulk/ .
initializeUnorderedBulkOp now not get options, but method saveBatch wait options. I think better solution delete this parameter "options" from saveBatch in new version (0.8). But now using my way.


brianhyder:
This looks good!  Thank you for the fix and contribution.

Issue #831, opened 2015-11-11
Enhancement - Limit functionality article editor (text)


adrianeguez:
### Text
It will be awesome that you can customize colors, 
![image](https://cloud.githubusercontent.com/assets/5687655/11106633/96082716-88a5-11e5-89dc-db9228d3557d.png)

blakecallens:
We'll look into adding this. In yhe meantime, you can change the colors in the HTML editor view.

webdesus:
I will do this on the this week.

PR #1155, opened 2016-10-04
Fixed bugs wysiwyg

webdesus:
1. Fixed problem with forever deselect text. I think this problem can be seen only on windows os. If you selected text then deselect text might only to switch on another html element.
2. Removed $interval from wysiwyg.js. I think watch will be enough.
3. Fixed problem with lost focus if click on modal dialog img. As a result image not added.
4. Opportunity using many wysiwyg on the form. And fixed bug with save select position cursor.

### DIFF
--- a/plugins/pencilblue/templates/admin/elements/wysiwyg_directive.html
+++ b/plugins/pencilblue/templates/admin/elements/wysiwyg_directive.html
@@ -1,7 +1,7 @@
 <div>
   <div class="wysiwyg_buttons well well-sm" style="margin-bottom: 0">
     <div class="btn-toolbar" role="toolbar">
-      <div class="btn-group btn-group-sm btn-group-format">
+      <div class="btn-group btn-group-sm">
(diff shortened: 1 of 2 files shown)


blakecallens:
@webdesus did you test this on Mac or Linux? Also which browsers?

webdesus:
I tested it on Linux(chrome) and Windows(chrome). I not had opportunity to test on mac, but mac have same event model like linux. I think all will be correct. Now I tested in firefox and ie. My code work perfect, but in IE wysiwyg has problem which had and previously([details](http://stackoverflow.com/questions/10741831/execcommand-formatblock-headings-in-ie)). I can fix it, if you want. 

I some days work with this code and found bug. Now I add commit to this pull request.

webdesus:
I fixed problem with formatblock in Internet Explorer, but IE can not work with blockquote mode

webdesus:
I realize feature #831. And I want add more functional for image in wysiwyg. How make better, wait when this pull request will merge with master and then add new pull request, or add to this pull request feater #831?

Issue #980, opened 2016-03-25
Crash opening article admin view as writer


dennisgermany:
Hi,

I'm getting the following error when opening the article admin view with role WRITER. (The other roles seem to work fine.)

ReferenceError: where is not defined
    at ManageArticles.render (/Users/dennis/Development/git/pencilblue/plugins/pencilblue/controllers/admin/content/articles/manage_articles.js:36:13)
    at /Users/dennis/Development/git/pencilblue/include/http/request_handler.js:1148:79
    at process._tickDomainCallback (node.js:381:11)

I just tried it again with the current master 0.5.1 and the error is still there.

Thanks!
--Dennis

brianhyder:
Confirmed as a bug.  The bug was introduced when the controller was switched over to retrieve the data via the API and not render as part of the page.  The original behavior was for the view to only render the articles created by the writer.  This is probably not intended behavior.  A writer should be able to view all articles but should not be able to edit all articles.

The fix should ensure:
* the article service validation checks the permission levels
* the article API allows for filtering by author

brianhyder:
UPDATE: The ability to filter articles by author will be addressed in #1006

ceballos59:
I have a cuestion ? - In current version Penciblue writers can not yet manage articles ?

brianhyder:
We are planning to correct this as part of the 1.0 release.  We are happy to entertain PRs to fix it in the meantime

PR #1173, opened 2016-11-17
Fixes linting issues

dukeofsussex:
Fixes #

**Description:**
Start fixing those bithound complaints (going to have to be done sooner or later)

@pencilblue/owners

@brianhyder Please check the second commit, needs thorough reviewing.

### DIFF
--- a/include/service/json_fs_entity_service.js
+++ b/include/service/json_fs_entity_service.js
@@ -58,7 +58,7 @@ module.exports = function JSONFSEntityServiceModule(pb) {
             catch(e) {
                 var error = util.format("%s: Failed to parse JSON from file: %s", this.type, key);
                 pb.log.error(error);
-                cb(new PBError(error).setSource(e));
+                cb(new pb.PBError(error).setSource(e)); // PBError class necessary?
             }
         };
         JSONFSEntityService.super_.prototype.render.apply([this, key, handler]);
@@ -74,7 +74,7 @@ module.exports = function JSONFSEntityServiceModule(pb) {
      */
     JSONFSEntityService.prototype.set = function(key, value, cb) {
         if (!util.isObject(value) && !util.isArray(value)) {
-            cb(new PBError(this.type+": Value must be an array or object: "+util.inspect(value)), null);
+            cb(new pb.PBError(this.type+": Value must be an array or object: "+util.inspect(value)), null); // PBError class necessary?
         }
 
         try {
--- a/include/service/media/renderers/daily_motion_media_renderer.js
+++ b/include/service/media/renderers/daily_motion_media_renderer.js
@@ -164,7 +164,7 @@ module.exports = function AudioMediaRendererModule(pb) {
      */
     DailyMotionMediaRenderer.getType = function(urlStr) {
         return DailyMotionMediaRenderer.isSupported(urlStr) ? TYPE : null;
-    }
+    };
 
     /**
      * Retrieves the Font Awesome icon class.  It is safe to assume that the type
--- a/include/service/media/renderers/instagram_media_renderer.js
+++ b/include/service/media/renderers/instagram_media_renderer.js
@@ -144,7 +144,7 @@ module.exports = function InstagramMediaRendererModule(pb) {
      */
     InstagramMediaRenderer.getType = function(urlStr) {
         return InstagramMediaRenderer.isSupported(urlStr) ? TYPE : null;
-    }
+    };
 
     /**
      * Retrieves the Font Awesome icon class.  It is safe to assume that the type
--- a/include/service/media/renderers/kickstarter_media_renderer.js
+++ b/include/service/media/renderers/kickstarter_media_renderer.js
@@ -51,7 +51,7 @@ module.exports = function KickStarterMediaRendererModule(pb) {
     var STYLES = Object.freeze({
 
         view: {
-            width: "100%",
+            width: "100%"
         },
 
         editor: {
@@ -144,7 +144,7 @@ module.exports = function KickStarterMediaRendererModule(pb) {
      */
     KickStarterMediaRenderer.getType = function(urlStr) {
         return KickStarterMediaRenderer.isSupported(urlStr) ? TYPE : null;
-    }
+    };
 
     /**
      * Retrieves the Font Awesome icon class.  It is safe to assume that the type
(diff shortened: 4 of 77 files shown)


dukeofsussex:
Bithound not complaining? How odd

include/service/media/renderers/daily_motion_media_renderer.js
```diff
@@ -149,7 +149,7 @@ module.exports = function AudioMediaRendererModule(pb) {
      * @return {Boolean} TRUE if URL points to the main domain and media resource, FALSE if not
      */
     DailyMotionMediaRenderer.isLibyanDomain = function(parsedUrl) {
-        if (util.isString(parsedUrl)) {
+        if (util.isString(parsedUrl)) { // Necessary?
```
brianhyder: if I recall correctly these were necessary based on some code paths.  Please review the comment and the other for this file.

include/service/json_fs_entity_service.js
```diff
@@ -74,7 +74,7 @@ module.exports = function JSONFSEntityServiceModule(pb) {
      */
     JSONFSEntityService.prototype.set = function(key, value, cb) {
         if (!util.isObject(value) && !util.isArray(value)) {
-            cb(new PBError(this.type+": Value must be an array or object: "+util.inspect(value)), null);
+            cb(new pb.PBError(this.type+": Value must be an array or object: "+util.inspect(value)), null); // PBError class necessary?
```
brianhyder: I do not believe it is.  At one point, early on, there was special logic to handle these errors.  That is no longer the case.  However, without tests there is not way to no for sure.  I would just leave the comments in for now.

include/service/media/renderers/instagram_media_renderer.js
```diff
@@ -129,7 +129,7 @@ module.exports = function InstagramMediaRendererModule(pb) {
      * @return {Boolean} TRUE if URL points to the main domain and media resource, FALSE if not
      */
     InstagramMediaRenderer.isFullSite = function(parsedUrl) {
-        if (util.isString(parsedUrl)) {
+        if (util.isString(parsedUrl)) { // Necessary?
```
brianhyder: please remove the comment

include/service/media/renderers/kickstarter_media_renderer.js
```diff
@@ -129,7 +129,7 @@ module.exports = function KickStarterMediaRendererModule(pb) {
      * @return {Boolean} TRUE if URL points to the main domain and media resource, FALSE if not
      */
     KickStarterMediaRenderer.isFullSite = function(parsedUrl) {
-        if (util.isString(parsedUrl)) {
+        if (util.isString(parsedUrl)) { // Necessary?
```
brianhyder: please remove the comment

include/service/media/renderers/daily_motion_media_renderer.js
dukeofsussex: What has me confused is that a parsedUrl is being parsed again

dukeofsussex:
@brianhyder I was going to remove all comments, most are just there to clarify stuff. And all the "Necessary?" ones are duplicates, as the corresponding code snippet has been copied to every renderer and semantically doesn't make sense

plugins/pencilblue/controllers/admin/content/articles/manage_articles.js
```diff
@@ -34,7 +34,7 @@ module.exports = function(pb) {
     ManageArticles.prototype.render = function(cb) {
         var self = this;
         if(!pb.security.isAuthorized(this.session, {logged_in: true, admin_level: pb.SecurityService.ACCESS_EDITOR})) {
-            where.author = this.session.authentication.user_id;
+            where.author = this.session.authentication.user_id; // Where is the object? Is this just a copy-paste?
```
brianhyder: This is set when the user logs in

dukeofsussex: This one actually gave me an error today, stating `where is not defined` when viewing the exact page as a writer

plugins/portfolio/controllers/blog.js
```diff
@@ -394,7 +394,7 @@ module.exports = function BlogModule(pb) {
         if(this.req.pencilblue_article || this.req.pencilblue_page) {
             cb(null, content.headline + ' | ' + pb.config.siteName);
         }
-        else if(searchId = this.req.pencilblue_section || this.req.pencilblue_topic) {
+        else if(searchId = this.req.pencilblue_section || this.req.pencilblue_topic) { // Is this an assignment during evaluation?
```
brianhyder: yes, the order of operations coalesces the section and topic before assigning them to the search id variable

brianhyder:
For the most part this looks great.  Just need to remove a few inline comments and then i'll merge.

dukeofsussex:
I'll tidy it up and let you know when it's ready

dukeofsussex:
@brianhyder Is include/service/entities/theme_service.js used anywhere? Couldn't find anything

plugins/pencilblue/controllers/admin/content/articles/manage_articles.js
brianhyder: Oh, I see.  This is related to #980.

include/service/media/renderers/daily_motion_media_renderer.js
dukeofsussex: Okay, I'll leave it be. Just seems redundant, but it's only a small check.

brianhyder:
The theme_service.js file may not be.  There are a few other unused files too but we are waiting until the next version to rip it out so that we don't break anything that other users were relying on.

plugins/pencilblue/controllers/admin/content/articles/manage_articles.js
dukeofsussex: Ah right. Just going to leave it then, too bad bithound :smile:

dukeofsussex:
Should be good to go, already rebased onto 0.8.0 for an easy merge (I hope 😉 )

plugins/pencilblue/controllers/preview.js
```diff
@@ -56,10 +55,6 @@ module.exports = function PreviewModule(pb) {
         });
     };
 
-    Preview.prototype.getPageTitle = function() {
-        return article.name;
-    };
```
dukeofsussex: The case was redundant,  as it uses the same as the default case, so why have it? (and bithound didn't like the syntax for whatever reason) And if I'm not mistaken this function isn't

PR #1176, opened 2016-11-18
Fixes #1174, #1175 and #1177

### LINKED ISSUE #1174: Duplicate Error when clicking on Forgot password Multiple Times
**Description:**
Clicking on the forgot password twice attempts to reinsert the password reset, with the same user id giving a duplicate error. Also, I think the password reset on a multitenancy site should be coming from the global site's email settings. Otherwise if the subdomain did not do the email setup, the reset password email would not be able to send

**Steps to Reproduce:**
On a subdomain, click on Forgot Password http://subdomain.localhost:8080/admin/login

**Configuration Details:**
Multitenancy is on

**Log Output:**
There was an error saving: E11000 duplicate key error index: mydb.password_reset.$userId_1 dup key: { : "582ddc743a089b2e70d19aab" }

PencilBlue Version: 
0.7.0 with multitenancy
Node.js Version:
4.2
Environment Description:

### LINKED ISSUE #1175: Cannot send a proper password reset from the admin screen of a multitenant subdomain
**Description:**
If you are on the admin page of a subdomain, sending a password reset request, does not specify the site property in the password_reset collection. The email that is sent to the user is a naked domain, ie localhost:8080 instead of subdomain.localhost:8080. This is a different action than the subdomain.localhost/user/login, where reset does specify the site field and a proper email verification link.

**Steps to Reproduce:**
On a subdomain, click on a user and send password reset request: 

http://subdomain.localhost:8080/actions/admin/users/send_password_reset/582ddc743a089b2e70d19aab

**Configuration Details:**
Multitenant Site

**Log Output:**

PencilBlue Version: 
0.7.0
Node.js Version:
0.4
Environment Description:

dukeofsussex:
Fixes #1175 

**Description:**
Fixes an issue with the password reset email not containing the correct subdomain and verification code

@pencilblue/owners

### DIFF
--- a/include/service/entities/password_reset_service.js
+++ b/include/service/entities/password_reset_service.js
@@ -96,7 +96,7 @@ module.exports = function(pb) {
             }
 
             //attempt to retrieve any existing reset
-            self.getSingle({where: {userId: userObj.id}}, function(err, passwordResetObj) {
+            self.getSingle({where: {userId: userObj[pb.DAO.getIdField()].toString()}}, function(err, passwordResetObj) {
                 if (util.isError(err)) {
                     return cb(err);
                 }
@@ -104,7 +104,7 @@ module.exports = function(pb) {
                 //need to know if we should create the DTO or not
                 var created = !passwordResetObj;
                 if (created) {
-                    passwordResetObj = {userId: userObj.id};
+                    passwordResetObj = {userId: userObj[pb.DAO.getIdField()].toString()};
                 }
 
                 //now persist it back
--- a/plugins/pencilblue/controllers/actions/admin/users/send_password_reset.js
+++ b/plugins/pencilblue/controllers/actions/admin/users/send_password_reset.js
@@ -49,7 +49,7 @@ module.exports = function(pb) {
              * @property dao
              * @type {DAO}
              */
-            self.dao = new DAO();
+            self.dao = new pb.SiteQueryService(self.getServiceContext());
 
             cb(err, true);
         };
@@ -79,10 +79,13 @@ module.exports = function(pb) {
                 }
 
                 if(!passwordReset) {
-                    passwordReset = pb.DocumentCreator.create('password_reset', {user_id: user[pb.DAO.getIdField()].toString()});
+                    passwordReset = {
+                        userId: user[pb.DAO.getIdField()].toString(),
+                        object_type: 'password_reset'
+                    };
                 }
 
-                passwordReset.verification_code = util.uniqueId();
+                passwordReset.verificationCode = util.uniqueId();
 
                 self.dao.save(passwordReset, function(err, result) {
                     if(util.isError(err)) {
--- a/plugins/pencilblue/controllers/actions/forgot_password.js
+++ b/plugins/pencilblue/controllers/actions/forgot_password.js
@@ -43,7 +43,9 @@ module.exports = function ForgotPasswordControllerModule(pb) {
         var ctx = this.getServiceContext();
         ctx.userService = this.userService;
         ctx.siteService = new SiteService(this.getServiceContext());
-        ctx.emailService = new EmailService(this.getServiceContext());
+        
+        // Use the global email settings in a multisite environment
+        ctx.emailService = new EmailService(util.union(this.getServiceContext(), {onlyThisSite: false}));
 
         /**
          * @property passwordResetService
--- a/plugins/pencilblue/controllers/actions/user/reset_password.js
+++ b/plugins/pencilblue/controllers/actions/user/reset_password.js
@@ -79,7 +79,7 @@ module.exports = function (pb) {
                 if (!data.user) {
                     self.formError(self.ls.g('users.INVALID_VERIFICATION'), '/user/login', cb);
                 }
-                self.passwordResetService.getSingle({where: {userId: data.user.id + '', verificationCode: self.query.code}}, callback);
+                self.passwordResetService.getSingle({where: {userId: data.user[pb.DAO.getIdField()].toString(), verificationCode: self.query.code}}, callback);
             }],
             deletePasswordReset: ['user', 'passwordReset', function(callback, data) {
 
--- a/plugins/pencilblue/controllers/actions/user/sign_up.js
+++ b/plugins/pencilblue/controllers/actions/user/sign_up.js
@@ -104,6 +104,8 @@ module.exports = function (pb) {
                 post.verification_code = util.uniqueId();
             }
 
+            // This is the reason why usernames are tranformed to lowercase when signing up
+            // TODO: Change sign-up behaviour in 1.0
             var user = pb.DocumentCreator.create(collection, post);
 
             self.validateUniques(user, function(err, results) {
--- a/include/email.js
+++ b/include/email.js
@@ -33,7 +33,7 @@ module.exports = function EmailServiceModule(pb) {
      */
     function EmailService(options) {
         if (options) {
-            this.site = pb.SiteService.getCurrentSite(options.site) || pb.SiteService.GLOBAL_SITE;
+            this.site = pb.SiteService.getCurrentSite(options.site);
             this.onlyThisSite = options.onlyThisSite || false;
         }
     }
--- a/include/security/authentication/index.js
+++ b/include/security/authentication/index.js
@@ -92,9 +92,10 @@ module.exports = function AuthenticationModule(pb) {
             return cb(new Error("FormAuthentication: The postObj parameter must be an object: "+postObj), null);
         }
 
-        //call the parent function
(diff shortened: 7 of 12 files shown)


include/model/create_document.js
```diff
@@ -106,18 +105,6 @@ module.exports = function DocumentCreatorModule(pb) {
...
-     * Formats usernames
-     *
-     * @method usernameFormatting
-     * @param {Object} post Key value pair object
-     */
-    DocumentCreator.usernameFormatting = function(post){
-        if(util.isString(post['username'])) {
-            post['username'] = post['username'].toLowerCase();
-        }
-    };
```
brianhyder: Why were these removed?  We prefer not to remove code at this time because it causes breaking changes for our users.

dukeofsussex: Should I rather keep this and remove the call from the create() function?

include/service/entities/password_reset_service.js
```diff
@@ -96,15 +96,15 @@ module.exports = function(pb) {
             }
 
             //attempt to retrieve any existing reset
-            self.getSingle({where: {userId: userObj.id}}, function(err, passwordResetObj) {
+            self.getSingle({where: {user_id: userObj[pb.DAO.getIdField()].toString()}}, function(err, passwordResetObj) {
```
brianhyder: Why is this separated by an underscore?

plugins/pencilblue/controllers/actions/user/reset_password.js
```diff
@@ -79,15 +79,17 @@ module.exports = function (pb) {
                 if (!data.user) {
                     self.formError(self.ls.g('users.INVALID_VERIFICATION'), '/user/login', cb);
                 }
-                self.passwordResetService.getSingle({where: {userId: data.user.id + '', verificationCode: self.query.code}}, callback);
+                self.passwordResetService.getSingle({where: {user_id: data.user[pb.DAO.getIdField()].toString(), verification_code: self.query.code}}, function (err, data) {
+                    callback(err, data);
+                });
```
brianhyder: Why is this anonymous function, that wraps the callback, necessary?

include/service/entities/password_reset_service.js
dukeofsussex: Because the database is user_id, not userId?

plugins/pencilblue/controllers/actions/user/reset_password.js
dukeofsussex: Good catch. It's a remnant from debugging, will fix

include/service/entities/password_reset_service.js
brianhyder: When we recently changed this service the plan to was eliminate the underscore.  Was there one that we missed?

include/model/create_document.js
brianhyder: I guess I don't understand why the call needed to be removed from the create function

include/service/entities/password_reset_service.js
dukeofsussex: Oh snap, and I thought it was the other way round. Welp, this PR needs an overhaul then, seems like I was looking in the wrong place.

include/model/create_document.js
dukeofsussex: The long version is in #1177. TL;DR: Usernames shouldn't be converted to lowercase when a user signs up

brianhyder: Got it.  We plan to move the whole "DocumentCreator" prototype.  It doesn't serve much of a purpose.  I would suggest implementing the fix such that it doesn't rely on the document creator.  Two suggestions: either use the UserService to create the user or call the individual functions from the create method.

plugins/pencilblue/controllers/actions/forgot_password.js
```diff
@@ -43,7 +43,7 @@ module.exports = function ForgotPasswordControllerModule(pb) {
         var ctx = this.getServiceContext();
         ctx.userService = this.userService;
         ctx.siteService = new SiteService(this.getServiceContext());
-        ctx.emailService = new EmailService(this.getServiceContext());
+        ctx.emailService = new EmailService({site: ctx.site});
```
brianhyder: Why change this?  The reason we had it to pull the whole service context was so that if the EmailService expanded to need more parameters we wouldn't have to change code where the service was instantiated.

dukeofsussex: This was changed to the global email settings are used and not the site specific ones. I can just as well make a copy of the context and set `onlyThisSite` to false (I think)

brianhyder:
Possibly, last question:  What pieces have you tested after changing?  I want to make sure we cover all of the scenarios before merging.

dukeofsussex:
@brianhyder I've tested the first 3 issue fixing commits on multisite, the last commit only changed the naming convention from underscore to camelCase, I didn't actually test the complete logic (I'm guessing it's broken as it was before the change)

plugins/pencilblue/controllers/actions/admin/users/send_password_reset.js
```diff
@@ -79,10 +79,14 @@ module.exports = function(pb) {
...
-                    passwordReset = pb.DocumentCreator.create('password_reset', {user_id: user[pb.DAO.getIdField()].toString()});
+                    passwordReset = {
+                        userId: user[pb.DAO.getIdField()].toString(),
+                        object_type: 'password_reset'
+                    };
                 }
 
-                passwordReset.verification_code = util.uniqueId();
+                passwordReset.verificationCode = util.uniqueId();
+                passwordReset.site = self.context.site;
```
brianhyder: Instead of manually setting the site, you can change line 52 such that: ```this.dao = new SiteQueryService(self.getServiceContext());```.  This will ensure that the site is automatically injected when the object is saved.

plugins/pencilblue/controllers/actions/forgot_password.js
```diff
@@ -41,9 +41,10 @@ module.exports = function ForgotPasswordControllerModule(pb) {
         this.userService = new UserService(this.getServiceContext());
 
         var ctx = this.getServiceContext();
+        ctx.onlyThisSite = false; // Use the global email settings in a multisite environment
         ctx.userService = this.userService;
-        ctx.siteService = new SiteService(this.getServiceContext());
-        ctx.emailService = new EmailService(this.getServiceContext());
+        ctx.siteService = new SiteService(ctx);
+        ctx.emailService = new EmailService(ctx);
```
brianhyder: I'd like to ask that we don't make the change to use the same context object.  The reason is because at some point down the road the context object could be modified and effect one or more of the services' behaviors.  Providing each instance with its own service context prevents bad things from happening if one service misbehaves.

plugins/pencilblue/controllers/actions/user/sign_up.js
```diff
@@ -94,59 +95,31 @@ module.exports = function (pb) {
...
                 successRedirect = '/user/verification_sent';
                 successMsg      = self.ls.g('users.VERIFICATION_SENT') + post.email;
-                post.verification_code = util.uniqueId();
+                post.verificationCode = util.uniqueId();
             }
 
-            var user = pb.DocumentCreator.create(collection, post);
-
-            self.validateUniques(user, function(err, results) {
+            self.service.add(post, function(err, data) {
```
brianhyder: We can't use the user service here.  The reason is because of a poor design decision we made early on with PB.  Unverified users are not in the same mongo collection as the rest of the users.  The user service only deals with verified (activated) users.  This is why the collection variable was used in the previous implementation.  It determined which collection to put the object in at runtime based on the content settings.  

We should revert this change.  The next version of PB (1.0) will make this breaking change so that we can simplify the logic as you have done by using the user service.

brianhyder:
I want to apologize for taking so long to look this over.  After diving in there are still a few changes that need to be made.  The hard part about the user management is that usernames and email addresses should be case insensitive.  In your use case, ```hello``` should be equivalent to ```Hello``` no matter whether you are logging in, resetting a password, or searching for a user.

dukeofsussex:
I agree and disagree at the same time. I agree that names and email addresses should be unique, but when a user creates an account with the name `MyName`, then I'd expect the name to be just that and not lowercased as it is now (even though this is only done by the DocumentCreator atm, so I'm assuming it will be removed later on).

brianhyder:
Yes, that is correct.  To clarify, I would expect that PB would persist the username and email as entered but when searching, logging in or, or validating they should be looked at as case insensitive.

dukeofsussex:
Sounds good 👍

dukeofsussex:
Okay, I've applied the requested changes. Hope I didn't miss anything else

brianhyder:
This looks good.  I'm going to go through and do some smoke testing tonight.  My hope is to have it merged in by this evening

PR #1183, opened 2016-12-02
Fixes plugin dialog modal behaviour

dukeofsussex:
Fixes the progress modal closing down early and not showing the whole job log

**Description:**
TL;DR I messed up 😞 

@pencilblue/owners

### DIFF
--- a/include/service/jobs/job_runner.js
+++ b/include/service/jobs/job_runner.js
@@ -237,7 +237,7 @@ module.exports = function JobRunnerModule(pb) {
 
         var query   = pb.DAO.getIdWhere(this.getId());
         var updates = {};
-        if (pb.validation.isNum(progressIncrement, true) && progressIncrement > 0) {
+        if (pb.validation.isNum(progressIncrement, true) && progressIncrement >= 0) {
             updates.$inc = {progress: progressIncrement};
         }
         if (pb.validation.isNonEmptyStr(status, true)) {


brianhyder:
Why would the greater than our equal too make a difference?

dukeofsussex:
What equal to?

brianhyder:
I don't understand how the change fixes the problem. Can you explain a
little bit?

dukeofsussex:
Sure, so here's what was going wrong:

Backend:

- Trying to install plugin
- `isInstalled` check passes fine
- `PluginAvailableJob`
    - https://github.com/pencilblue/pencilblue/blob/master/include/service/jobs/plugins/plugin_available_job.js#L51
        - Calls `self.onUpdate(increment)` with `increment = 0`
    - https://github.com/pencilblue/pencilblue/blob/master/include/service/jobs/job_runner.js#L235
        - Increment = 0, so our `updates` object is still empty
        - Status is empty, so our `updates` object is still empty
    - https://github.com/pencilblue/pencilblue/blob/master/include/service/jobs/job_runner.js#L248
        - `updates` is empty, so the job store is **not** updated
        
-> `job_run` progress and status are not initialized

Frontend:

- The progress modal asks for the current job status
- https://github.com/pencilblue/pencilblue/blob/master/plugins/pencilblue/templates/angular/admin/plugins/manage_plugins.html#L81
    - `progress` doesn't exist, so `undefined` is printed to the log window
- https://github.com/pencilblue/pencilblue/blob/master/plugins/pencilblue/templates/angular/admin/plugins/manage_plugins.html#L85
    - `status` doesn't exist either, so the `else` path is taken
    
-> Modal is closed down early

Hope this sort of makes sense.

PR #1185, opened 2016-12-05
Fixes #970 - Improved Routing Capabilities

### LINKED ISSUE #970: Request Handler Refactor
The refactor will attempt to breakup the request_handler to be more modular.  it will focus on being more "express" like where handlers can be registered in various parts of the pipeline.  The hope is to begin to support plugins that provide specific functionality such as security and principal population

brianhyder:
Fixes #970 

**Description:**
Starts the extraction of the routing framework from the request handling.  It creates a formal pipeline that plugins can tap into in a variety of ways by adding, removing, or replacing middleware.

I will be performing further testing but wanted to get the PR up for review.

@pencilblue/owners


include/http/middleware/index.js
```diff
@@ -0,0 +1,346 @@
...
+            catch (e) {
+                pb.log.error('RequestHandler: Failed to set cookie: %s', e.stack);
+            }
+        }
+        next();
+    };
+
+    Middleware.writeResponse = function (req, res, next) {
+        var data = req.controllerResult;
+        var doRedirect = typeof data.redirect !== 'undefined';
```
dukeofsussex: Is this variable necessary, as it's not reused and the check could just as well be done in the if-statement itself?

PR #1196, opened 2017-01-02
1.0.0

MarcosRava:
Adapt code to ecma6 bases on airbnb eslint


MarcosRava:
@dukeofsussex Done!

MarcosRava:
@dukeofsussex Done! I think ._.

dukeofsussex:
Another small typo in https://github.com/pencilblue/pencilblue/pull/1196/commits/337a31e7cc21965efda38243ebf3cf4f4581a213 😆 

I think you now also changed the commit order, was that intended?

dukeofsussex:
Thanks for opening the PR, I had completely forgotten to open one for https://github.com/dukeofsussex/pencilblue/commit/19ca0780e115c077ff241c94801d979f336d93d8

MarcosRava:
My mistake about commit order -.-'

About https://github.com/dukeofsussex/pencilblue/commit/19ca0780e115c077ff241c94801d979f336d93d8 
do you want that I adapt the changes to this eslint.js?

dukeofsussex:
@MarcosRava Not sure, best to wait for @brianhyder's verdict

MarcosRava:
@dukeofsussex ok, tks!

brianhyder:
Haven't had a chance to look at the PR yet. I'll have an answer for you by
tomorrow evening.

controllers/api/base_api_controller.js
```diff
@@ -15,56 +15,135 @@
...
+
+    constructor() {
+      super();
+      /**
+       * Processes the query string to develop the where clause for the query request
+       * @method processWhere
+       * @param {Object} q The hash of all query parameters from the request
+       * @return {Object}
+       */
+      this.processWhere = function processWhere() {
```
brianhyder: We would prefer to keep the constructor simple and clean.  Declaring the instance functions in the constructor nests them deeper than necessary and makes testing harder because you can't stub the prototype before the instance is created.

brianhyder:
As for your question about the linters.  That isn't my area of expertise.  Whatever you all think is the best way to go I'm ok with just need someone to provide a couple of justifications so that when other users ask we can tell them.  

Lastly, I'd like to leave the indentation for tabs at 4 spaces.  This might seem silly but we use it a deterrent for writing deeply nested code.  If it is getting to the point where it is leaning a bit to the right then it is time to start extracting things into smaller functions that are not deeply nested.  This also helps with testability because it is easier to stub those pieces out individually.

.eslintignore
```diff
@@ -0,0 +1,4 @@
+test/*
+public/*
+coverage/*
+node_modules/*
```
MarcosRava: I didn't put all the things, because I didn't know if the proposal would be accepted

.editorconfig
```diff
@@ -2,11 +2,11 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
MarcosRava: Yes, I changed to be compatible with airbnb linter

controllers/api/base_api_controller.js
MarcosRava: There are many instance functions that not use the instance (this), I put them inside the constructor because this rule http://eslint.org/docs/rules/class-methods-use-this, I think that this functions should be statics
I didn't want to do a deeper refactoring because I still don't understand all pencilblue ecosystem

MarcosRava:
@brianhyder about linter justifications, AirBnB has all justifications about rules here https://github.com/airbnb/javascript, I decided to use AirBnB linter because many communities and projects adhered to them

controllers/api/base_api_controller.js
brianhyder: Sounds good to me.

brianhyder:
@MarcosRava, @dukeofsussex This looks good y'all.  Just let me know when you are ready to merge and I'll click the button.

MarcosRava:
@dukeofsussex Is it ok for you?

dukeofsussex:
Could you give me a few days? I'd love to go through their ruleset, as they seem pretty similar to my suggested ones.

Only issue I have depending on somebody else's linter is that any changes made by them might not be deemed necessary in PB and might cause issues later on 😒

PR #1265, opened 2020-02-16
changing variable types to const instead of var

noobmaster440:
Fixes #

**Description:**
changing variable types to const instead of var as per es6 rules

@pencilblue/owners


noobmaster440:
Thank you so much Brian. It means a lot to me.
Really appreciate it

